PR #1, opened 2022-03-04
Add ci / linter workflow

marcushines:
This adds go linting and go test / build workflows

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,32 @@
+name: Go
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2.1.3
+      with:
+        go-version: 1.x
+    - name: Build
+      run: go build -v ./...
+  test:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2.1.3
+      with:
+        go-version: 1.16.7
```
robshakir, line 17: Is this specific version required? We've generally found it a bit better just to have `go-version: 1.x` here I think in order to avoid having to add new versions.

.github/workflows/go.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Build
+      run: go build -v ./...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Go
+        uses: actions/setup-go@v2.1.3
+        with:
+          go-version: 1.17.7
```
robshakir, line 27: Same question here as above. Should we have this on the same version as above?

robshakir (changes requested):

.github/workflows/go.yml
marcushines: done

.github/workflows/go.yml
marcushines: done

robshakir (approved):

PR #4, opened 2022-03-08
Add Generated GoStructs and PathStructs.

wenovus:
ONDATRA's idiosyncratic GoStructs are not suitable for use with the reference implementation, because it prefers state over config, and uses shadow paths, which is a feature that has limited support, which in turn makes working with it awkward.

Added `generate.sh` which is invokable via `go generate`. This file is adapted from ONDATRA's `internal/gnmigen/generate.sh`.


wenovus:
I think it's ok -- we'll just need to update the models and our logic for them. I think the evolution of the fake's models should be independent of Ondatra's own models changing. If they changed that shouldn't break us, but we can strive to keep them in lock-step.

marcushines (approved):

PR #7, opened 2022-04-06
update lemming with new call pattern and clean up couple nits

marcushines:


lemming.go
```diff
@@ -26,26 +25,26 @@ import (
 	fp4rt "github.com/openconfig/lemming/p4rt"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
+	"k8s.io/klog"
```
wenovus, line 28: "k8s.io/klog/v2"? Also should InitFlags(nil) be called? https://github.com/kubernetes/klog#how-to-use-klog However if we do call it I'm wondering if we're running into this problem: https://github.com/kubernetes/klog#coexisting-with-glog since other parts of the code base currently uses glog. So I'm wondering if we should convert to klog, if we choose to do so, all at once in a separate PR?

lemming.go
```diff
@@ -26,26 +25,26 @@ import (
...
 	s           *grpc.Server
-	addr        string
 	lis         net.Listener
 	stop        func()
 	gnmiServer  *fgnmi.Server
 	gnoiServer  *fgnoi.Server
 	gribiServer *fgribi.Server
 	gnsiServer  *fgnsi.Server
 	p4rtServer  *fp4rt.Server
+	err         error
```
wenovus, line 41: comment on what this error is and how it is used?

lemming.go
```diff
@@ -55,36 +54,35 @@ func New(lis net.Listener, opts ...grpc.ServerOption) (*Device, error) {
...
-	return d.addr
+	return d.lis.Addr().String()
 }
 
 // Stop stops the listening services.
-func (d *Device) Stop() {
-	if d.stop == nil {
-		return
-	}
+func (d *Device) Stop() error {
```
wenovus, line 67: suggestion to update doc comment to document where the error comes from.

lemming.go
```diff
@@ -55,36 +54,35 @@ func New(lis net.Listener, opts ...grpc.ServerOption) (*Device, error) {
...
+	go func() {
+		d.err = d.s.Serve(d.lis)
+	}()
 
 	d.stop = func() {
 		d.s.Stop()
-		d.lis.Close()
+		if err := d.lis.Close(); err != nil {
+			klog.Info(err)
+		}
```
wenovus, line 86: curious why log the error instead of storing it?

lemming.go
marcushines: we should switch all our code to use one and i would vastly recommend klog for everything

lemming.go
marcushines: because we don't really care about why closing the listener failed vs. the server actually failing

wenovus (approved):

PR #11, opened 2022-08-04
Refactor gnmit.New to allow gnmit to be bound to existing server.

robshakir:
```
* (M) gnmi/gnmit/gnmit.go
    - Modify gnmit to not create a new gRPC server, allowing it to
      be bound to an existing server.
```

The current gnmit implementation assumes that it can own the gNMI server, which
means it cannot be easily bound to a different gRPC server that already exists.
This PR creates a `NewServer` mechanism that can create the server and return it
such that it can be run on an existing port.

### DIFF
--- a/gnmi/gnmit/gnmit.go
+++ b/gnmi/gnmit/gnmit.go
@@ -209,8 +209,8 @@ func NewSettable(ctx context.Context, addr string, hostname string, sendMeta boo
 	}
 
 	vr, ok := schema.Root.(ygot.ValidatedGoStruct)
-	if err != nil {
-		return nil, "", fmt.Errorf("invalid schema root, %v", ok)
+	if !ok {
+		return nil, "", fmt.Errorf("invalid schema root, %v", schema.Root)
 	}
 	// Initialize the root with default values.
 	schema.Root.(populateDefaultser).PopulateDefaults()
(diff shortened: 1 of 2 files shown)


robshakir:
can I raise my objection to this linter again? this suggestion is pretty poor - it's basically failing because `mustTargetPath` (in tests) is used with a single name (`"local"`) for the target. this function does something useful - i.e., make tests significantly more readable.

changing the argument just to make the linter happy seems suboptimal. hardcoding the value into the helper function also feels a bad idea... is there a reason we need something more opinionated than the base linters we're using in other projects?

wenovus (approved):
Thanks for the improvement

wenovus:
> can I raise my objection to this linter again? this suggestion is pretty poor - it's basically failing because `mustTargetPath` (in tests) is used with a single name (`"local"`) for the target. this function does something useful - i.e., make tests significantly more readable.
> 
> changing the argument just to make the linter happy seems suboptimal. hardcoding the value into the helper function also feels a bad idea... is there a reason we need something more opinionated than the base linters we're using in other projects?

I agree with this sentiment to replace golang-ci similar with what featureprofiles has for `static_analysis` (which has `revive` on top of what ygot currently has). https://github.com/openconfig/featureprofiles/blob/main/.github/workflows/pull_request.yml#L117, https://github.com/openconfig/ygot/blob/master/.github/workflows/go.yml#L125

@marcushines what do you think of this direction as the common linter for all openconfig projects? Although it's not clear to me that there will be a single file to enforce this since each repo has its own idiosyncrasies (see ygot and featureprofiles).

PR #15, opened 2022-08-24
Add dataplane proto

DanG100:


proto/dataplane/dataplane.proto
```diff
@@ -0,0 +1,133 @@
...
+  uint64 prefix_len = 3;
+}
+
+message SubscribePortStateResponse {
+  string name = 1;
+  PortState admin_state = 2;
+  repeated IPNetwork network = 3;
+}
+
+message RouteDst {
```
robshakir, line 56: Consider calling this next-hop/NextHop?

proto/dataplane/dataplane.proto
```diff
@@ -0,0 +1,133 @@
...
+
+message InsertRouteRequest {
+  uint64 vrf = 1;
+  string prefix = 2;
+  repeated RouteDst dst = 3;
+}
+
+message InsertRouteResponse {}
+
+message DeleteRouteRequest {
```
robshakir, line 70: For modifying a route is the expectation that one calls delete and then add, or does add for an existing route replace one that is already there?

proto/dataplane/dataplane.proto
```diff
@@ -0,0 +1,133 @@
...
+  string prefix = 1;
+  uint64 vrf = 2;
+}
+
+message DeleteRouteResponse {}
+
+message AddNeighborRequest {
+  string ip = 1;
+  string hwaddr = 2;
+  bool transient = 3;
```
robshakir, line 80: Does transient here mean that this is an ARP entry that should time out? If so, did we consider maybe just having this be implemented outside of the dataplane? i.e., if an ARP entry is meant to timeout then the daemon that installed it should go remove it when it wants it to time out?

robshakir (approved):
LGTM, couple of comments :-)

proto/dataplane/dataplane.proto
DanG100: yeah, that was the intention. I'm in favor that being handled outside the dataplane

proto/dataplane/dataplane.proto
DanG100: done

proto/dataplane/dataplane.proto
DanG100: my idea was to call delete then add, but i'm don't have a strong opinion

wenovus: I think each route should be uniquely identified by a key, namely (prefix, vrf, address-family) as fields I think is enough. If so, then I would prefer "add" to mean a replace, since that sounds to me like the simplest semantic to give this operation in the presence of an existing route.

proto/dataplane/dataplane.proto
```diff
@@ -0,0 +1,132 @@
...
+}
+
+message SetPortStateResponse {}
+
+message SubscribePortStateRequest {}
+
+message IPNetwork {
+  string ip = 1;
+  uint64 prefix_len = 3;
+}
```
wenovus, line 48: (no action required) It looks like there is some overlap between our respective protos that we can consolidate later. I'll mark a TODO in my PR for this.

proto/dataplane/dataplane.proto
DanG100: updated comment to clarify behavior and allow overwriting

proto/dataplane/dataplane.proto
```diff
@@ -0,0 +1,133 @@
...
+  string name = 1;
+  string hwaddr = 2;
+  repeated IPNetwork ipv4 = 3;
+  repeated IPNetwork ipv6 = 4;
+}
+
+message UpdatePortResponse {}
+
+enum PortState {
+  PORTSTATE_INVALID = 0;
```
wenovus, line 31: nit: What's the purpose of INVALID here? I'm wondering because I'm thinking this might create extra handling code.

wenovus (approved):

proto/dataplane/dataplane.proto
DanG100: ack, a proto package with common types seems like a good idea

proto/dataplane/dataplane.proto
DanG100: the enum needs a default value and i wasn't convinced that DOWN is right as the default. I could rename this the PORTSTATE_UNSET if that's clearer

wenovus: I see, although I'm not sure what `PORTSTATE_UNSET` is the correct default either -- what does it even mean?

wenovus: Nvm, having an unset/unspecified value in an enum is in general good style, but if we ever add a new value that would cause trouble for backwards-compatibility.

PR #18, opened 2022-08-29
Add placeholder server, main, and dockerfile

DanG100:


dataplane/main/main.go
```diff
@@ -0,0 +1,51 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"net"
+
+	"github.com/openconfig/lemming/dataplane/server/server"
```
wenovus, line 22: Extra server folder?

dataplane/main/main.go
```diff
@@ -0,0 +1,51 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/local"
+	"google.golang.org/grpc/reflection"
+
+	log "github.com/golang/glog"
```
wenovus, line 27: Does klog work here? I didn't use it for the sysrib because I had some flag conflicts. (I will take a look before we integrate the pieces)

dataplane/main/main.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func main() {
+	flag.Parse()
+	addr := fmt.Sprintf("localhost:%d", *port)
+	list, err := net.Listen("tcp", addr)
+	if err != nil {
+		log.Fatalf("failed to listen on %q: %v", addr, err)
+	}
+	srv := grpc.NewServer(grpc.Creds(local.NewCredentials()))
```
wenovus, line 42: What does this do? Does the client need to authenticate in some way?

dataplane/main/main.go
DanG100: fixed

dataplane/main/main.go
DanG100: it's effectively the same as insecure.NewCredentials 
`grpc.Dial("localhost", grpc.WithTransportCredentials(local.NewCredentials())`

wenovus (approved):

PR #19, opened 2022-08-29
Example Declarative Configuration for establishing BGP session using GoBGP

wenovus:
goBgpTask's doc comments discusses the design for declarative configuration for GoBGP. It is similar in idea to GoBGP's implementation for declarative configuration using its own config language (instead of OpenConfig/gNMI):
https://github.com/osrg/gobgp/blob/a5bb236b0d8d2a976bea8c62ffb2ab5c70ed6e9d/pkg/config/config.go#L367

Unfortunately I can't get the connection to be established yet. It only goes to "ACTIVE".

Incidentally, GoBGP actually uses an old version of OpenConfig for its configuration, although it uses its own custom Go binding instead of ygot based on viper. We may want to use this when we're actually implementing GoBGP to avoid re-inventing the wheel. For other protocols we will need our own implementation since GoBGP's implementation is very BGP specific. See the "Assumptions:" section below to see my concerns for why we might want something different.


wenovus:
Workaround required to ignore staticcheck rules: https://github.com/golangci/golangci-lint/issues/741

gnmi/fakedevice/gobgp.go
```diff
@@ -0,0 +1,383 @@
...
+					shouldProcess = true
+				default:
+					log.V(1).Infof("goBgpTask: update path received isn't matched by any handlers: %s", prototext.Format(u.Path))
+				}
+			}
+			for _, u := range no.Delete {
+				log.V(1).Infof("Received delete path: %s", prototext.Format(u))
+				switch {
+				case len(u.Elem) > 0:
+				case len(u.Element) > 0: //nolint:staticcheck //lint:ignore SA1019 Specifically handling deprecated gNMI Element fields.
```
DanG100, line 201: i think we can not support about deprecated fields. we should return an error in RPCs. TODO is fine

gnmi/fakedevice/gobgp.go
```diff
@@ -0,0 +1,383 @@
...
+//     When there is a dependency, the later actions will NOT directly depend on the results of the previous actions, but will just look at the current config view to determine the appropriate action.
+//     e.g. for peers, if the global setting has been set up, then we can create the peers and update the applied config if it succeeds, but if not then we don't do anything.
+//     ; however, if the global setting hasn't been set up, we actually need to erase the entirety of the applied config. This is because the watcher doesn't tell us this information.
+func goBgpTask(getIntendedConfig func() *config.Device, q gnmit.Queue, update gnmit.UpdateFn, target string, remove func()) error {
+	bgpStatePath, _, err := ygot.ResolvePath(telemetrypath.DeviceRoot("").NetworkInstance("default").Protocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Bgp())
+	if err != nil {
+		log.Fatalf("goBgpTask failed to initialize due to error: %v", err)
```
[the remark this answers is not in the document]
wenovus: Done

gnmi/fakedevice/gobgp.go
```diff
@@ -0,0 +1,383 @@
...
+//     e.g. for Global if the intended config matches the applied config we will simply skip this step. If not we will actually do the action (either start, stop, or stop-start (aka. update)), and if it succeeds, update the applied config both in the DB as well as the current view of the applied config.
+//     When there is a dependency, the later actions will NOT directly depend on the results of the previous actions, but will just look at the current config view to determine the appropriate action.
+//     e.g. for peers, if the global setting has been set up, then we can create the peers and update the applied config if it succeeds, but if not then we don't do anything.
+//     ; however, if the global setting hasn't been set up, we actually need to erase the entirety of the applied config. This is because the watcher doesn't tell us this information.
+func goBgpTask(getIntendedConfig func() *config.Device, q gnmit.Queue, update gnmit.UpdateFn, target string, remove func()) error {
```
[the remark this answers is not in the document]
wenovus: Done.

gnmi/fakedevice/gobgp.go
wenovus: Done

gnmi/fakedevice/gobgp.go
```diff
@@ -0,0 +1,392 @@
...
+	bgpStatePath, _, err := ygot.ResolvePath(telemetrypath.DeviceRoot("").NetworkInstance("default").Protocol(telemetry.PolicyTypes_INSTALL_PROTOCOL_TYPE_BGP, "BGP").Bgp())
+	if err != nil {
+		log.Fatalf("goBgpTask failed to initialize due to error: %v", err)
```
DanG100, line 64: why are you logging and returning an error?

wenovus: Removed to avoid double logging.

DanG100 (approved):

PR #22, opened 2022-08-31
Integrate sysrib with dataplane

wenovus:
I didn't want to change testing since the dataplane API is still in
development. So I created another layer -- in order to keep the test
unchanged.

### DIFF
--- a/sysrib/server.go
+++ b/sysrib/server.go
@@ -26,13 +26,17 @@ import (
 	"github.com/openconfig/lemming/dataplane"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 
 	dpb "github.com/openconfig/lemming/proto/dataplane"
 	pb "github.com/openconfig/lemming/proto/sysrib"
 )
 
-const SockAddr = "/tmp/sysrib.api"
+const (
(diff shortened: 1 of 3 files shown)


sysrib/server.go
```diff
@@ -53,25 +60,57 @@ type Server struct {
...
+//
+// If dp is nil, then a connection attempt is made.
+func NewServer(dp dataplaneAPI) (*Server, error) {
 	rib, err := NewSysRIB(nil)
 	if err != nil {
 		return nil, err
 	}
 
+	if dp == nil {
+		opts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
```
DanG100, line 101: use local.NewCredentials()

DanG100 (approved):

sysrib/server.go
wenovus: Is this the yet newer way of specifying an insecure connection?

wenovus: Chatted offline, thanks.

wenovus: local.NewCredentials() is a ServerOption, and not a DialOption, so I can't quite use it. Since the current code follows the recommendation I'll keep it as is: https://pkg.go.dev/google.golang.org/grpc#WithInsecure, thanks.

DanG100: it's fine, but `grpc.Dial("", grpc.WithTransportCredentials(local.NewCredentials()))`

PR #23, opened 2022-08-31
UpdatePort implementation

DanG100:


dataplane/internal/kernel/kernel.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if up {
+		return netlink.LinkSetUp(link)
+	}
+	return netlink.LinkSetDown(link)
+}
+
+// CreateTAP creates kernel TAP interface.
+func CreateTAP(name string) error {
+	fd, err := unix.Open("/dev/net/tun", unix.O_RDWR, 0)
```
wenovus, line 76: (no action req) Why is it tun instead of tap? Not familiar with it, was just curious.

dataplane/server.go
```diff
@@ -24,3 +35,57 @@ const (
...
+	setInterfaceIPs    = kernel.SetInterfaceIPs
+	setInterfaceState  = kernel.SetInterfaceState
+)
+
+// UpdatePort updates an interface properties from the input request.
+func (s *Server) UpdatePort(_ context.Context, req *dpb.UpdatePortRequest) (*dpb.UpdatePortResponse, error) {
+	if req.Name == "" {
+		return nil, status.Error(codes.InvalidArgument, "name not set")
+	}
+	tapName := fmt.Sprintf("%s-tap", req.Name)
```
wenovus, line 51: Should this name conversion be a helper function?

dataplane/server.go
```diff
@@ -24,3 +35,57 @@ const (
...
+		})
+	}
+	if err := setInterfaceIPs(tapName, ips); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to set IPs: %v", err)
+	}
+	if req.AdminState != dpb.PortState_PORT_STATE_UNSPECIFIED {
+		if err := setInterfaceState(tapName, req.AdminState == dpb.PortState_PORT_STATE_UP); err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to set interface set: %v", err)
+		}
+	}
```
wenovus, line 89: switch statement in case we add another enum value?

dataplane/internal/kernel/kernel.go
DanG100: that's just the kernel calls it

dataplane/server.go
DanG100: will do once needed elsewhere

dataplane/server.go
DanG100: done

wenovus (approved):

PR #26, opened 2022-09-02
Fix ygnmi path structs

wenovus:
Need to use resolved pathstructs


DanG100:
thanks for fixing this. good catch i forgot about "resolved" path structs

DanG100 (approved):

PR #29, opened 2022-09-02
Add kernel port implementation

DanG100:


dataplane/forwarding/fwdport/ports/kernel.go
```diff
@@ -0,0 +1,174 @@
...
+	log "github.com/golang/glog"
+	fwdpb "github.com/openconfig/lemming/proto/forwarding"
+)
+
+func init() {
+	fwdport.Register(fwdpb.PortType_PORT_TYPE_KERNEL, kernelBuilder{})
+}
+
+type kernelBuilder struct {
+}
```
wenovus, line 38: a short comment please. I think my main confusion when starting to read this is what a "kernel" port is in the context of this code. From the proto it looks like it's a linux network device identified by a name. Would be nice to see a comment for this so it's clear from the outset.

dataplane/forwarding/fwdport/ports/kernel.go
```diff
@@ -0,0 +1,174 @@
...
+	if p.input, err = fwdaction.NewActions(kernelUpd.Kernel.GetInputs(), p.ctx); err != nil {
+		return fmt.Errorf("ports: input actions for port %v failed, err %v", p, err)
+	}
+	if p.output, err = fwdaction.NewActions(kernelUpd.Kernel.GetOutputs(), p.ctx); err != nil {
+		return fmt.Errorf("ports: output actions for port %v failed, err %v", p, err)
+	}
+	return nil
+}
+
+func (p *kernelPort) read() {
```
wenovus, line 118: Should this be named process() instead?

dataplane/forwarding/fwdport/ports/kernel.go
```diff
@@ -0,0 +1,174 @@
...
+	switch dir {
+	case fwdpb.PortAction_PORT_ACTION_INPUT:
+		return p.input
+	case fwdpb.PortAction_PORT_ACTION_OUTPUT:
+		return p.output
+	}
+	return nil
+}
+
+// State manages the state of the port.
```
wenovus, line 160: returns or manages? I guess it's not using `op` right now, so is there a TODO?

dataplane/forwarding/fwdport/ports/kernel.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (p *kernelPort) Cleanup() {
+	p.input.Cleanup()
+	p.output.Cleanup()
+	p.handle.Close()
+	p.input = nil
+	p.output = nil
+}
+
+// Update updates a port.
```
wenovus, line 95: updates what of a port?

dataplane/forwarding/fwdport/ports/kernel.go
DanG100: added todo

dataplane/forwarding/fwdport/ports/kernel.go
DanG100: sure

wenovus (approved):

PR #32, opened 2022-09-06
Misc cleanups to fwd.go

DanG100:
* implement  gRPC server
* remove logging

### DIFF
--- a/dataplane/forwarding/fwd.go
+++ b/dataplane/forwarding/fwd.go
@@ -85,33 +85,33 @@ func New(name string) *Engine {
 
 // client returns a gRPC client connected to the specified address.
 // It is assumed that clients are looked up while the service is locked.
-func (s *Engine) client(addr string) (*grpc.ClientConn, error) {
-	if c, ok := s.conn[addr]; ok {
+func (e *Engine) client(addr string) (*grpc.ClientConn, error) {
+	if c, ok := e.conn[addr]; ok {
 		return c, nil
 	}
 	c, err := grpc.Dial(addr)
 	if err != nil {
 		return nil, fmt.Errorf("service: dial to %v failed, err %v", addr, err)
 	}
-	s.conn[addr] = c
+	e.conn[addr] = c
 	return c, nil
 }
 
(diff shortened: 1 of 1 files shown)


dataplane/forwarding/fwd.go
```diff
@@ -124,54 +192,111 @@ func (f *Engine) UpdatePacketSink(contextID *fwdpb.ContextId, packet fwdcontext.
...
+			if err = e.UpdatePacketSink(cid, ps, paddr); err != nil {
+				return nil, err
+			}
+
+		case fwdpb.ContextUpdateRequest_OPERATION_UPDATE_NOTIFICATION_ADDRESS:
+			if err = e.UpdateNotification(cid, ns, naddr); err != nil {
+				return nil, err
+			}
+		}
+	}
```
wenovus, line 260: Should the second for loop just be merged with the first? It seems to be calling update multiple times so it's confusing what the intention of the first for loop is.

dataplane/forwarding/fwd.go
```diff
@@ -56,11 +60,18 @@ import (
 // forwarding contexts, each of which contain forwarding objects such as
 // tables, ports and actions.
 type Engine struct {
+	fwdpb.UnimplementedServiceServer
+	fwdpb.UnimplementedInfoServer
+	fwdpb.UnimplementedPacketSinkServer
+
```
wenovus, line 66: (no action required) it looks like you're now making this Engine implement the server where it did not before. I'm curious where this Engine was used prior to this refactoring?

dataplane/forwarding/fwd.go
DanG100: good catch

dataplane/forwarding/fwd.go
DanG100: there was a separate struct that implement in RPC interface and it called the Engine's  funcs, I just simplified it

wenovus: thanks for clarifying.

wenovus (approved):

PR #33, opened 2022-09-06
Prototype for a Task/Agent that is an external process to the datastore.

wenovus:
A wrapper is used to provide a modified gnmi.Set service. This gRPC service is currently provided via a unix socket.

Also refactored gnmit's Set. The new `set` function is what the new service calls.


proto/datastore/datastore.proto
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package datastore;
```
DanG100, line 17: why define a new RPC? couldn't we just use gnmi.Set instead of Update()

wenovus: In the future, we'd want gnmi.Set to error out whenever the user tries to write to a read-only value. Using a different RPC call allows read-only values to be over-written.

DanG100: sorry I wasn't clear.

What i was thinking was we'd could have a local (listening unix socket) gnmi server in addition to the "real" gnmi server. (2 total). Only the local Set implementation would allow writing to read-only paths. The local server doesn't have to implement subscribe, also the Set could only implement setting to simplify server. This would allow us to allow use the easy to use ygnmi.Update/Delete.

thoughts?

wenovus: Sounds good to me, I'll make the changes.

wenovus: Done.

gnmi/gnmit/gnmit.go
```diff
@@ -193,6 +192,24 @@ func New(ctx context.Context, addr, hostname string, sendMeta bool, tasks []Task
...
+	}
+	srvDS := grpc.NewServer(opts...)
+	gpb.RegisterGNMIServer(srvDS, NewDatastoreServer(gnmiserver))
+	lisDS, err := net.Listen("unix", DatastoreAddress)
+	if err != nil {
+		log.Fatalf("listen error: %v", err)
+	}
+	go func() {
+		if err := srvDS.Serve(lisDS); err != nil {
+			log.Errorf("Error while serving datastore target: %v", err)
```
DanG100, line 208: maybe make this log.Fatalf()? probably want to exit if the server stops

DanG100 (approved):

PR #35, opened 2022-09-07
Factor out creation of DatastoreService (alternate gNMI service)

wenovus:


gnmi/gnmit/gnmit.go
```diff
@@ -181,6 +181,27 @@ func NewServer(ctx context.Context, hostname string, sendMeta bool, tasks []Task
 	return c, gnmiserver, nil
 }
 
+func StartDatastoreServer(gnmiServer *GNMIServer) (func(), error) {
```
DanG100, line 184: add comment (especially clarifying return values

gnmi/gnmit/gnmit.go
```diff
@@ -181,6 +181,27 @@ func NewServer(ctx context.Context, hostname string, sendMeta bool, tasks []Task
...
+	if err := os.RemoveAll(DatastoreAddress); err != nil {
+		log.Fatal(err)
```
DanG100, line 186: return err instead

DanG100 (approved):

gnmi/gnmit/gnmit.go
wenovus: oops I put the comment in the subsequent PR.

gnmi/gnmit/gnmit.go
wenovus: Done.

PR #36, opened 2022-09-07
Attach gNMI service to lemming.go and create main package.

wenovus:


main/main.go
```diff
@@ -0,0 +1,31 @@
+package main
```
DanG100, line 1: optional: it's nice when the folder name is a meaningful name that way by default the built binary has a good name. there's an informal/unofficial project layout that I tend to reference https://github.com/golang-standards/project-layout

DanG100 (approved):

main/main.go
wenovus: Thanks, fixed.

PR #40, opened 2022-09-08
Move gnmi out of internal so it can accessed by dataplane

DanG100:


wenovus (changes requested):
LGTM pending lint config update to avoid scanning this dir.

wenovus (approved):

PR #45, opened 2022-09-13
Update state in interface handlers

DanG100:


gnmi/client/client.go
```diff
@@ -69,3 +69,39 @@ func NewYGNMIClient() (*ygnmi.Client, error) {
...
+}
+
+// BatchDelete stores an delete operation in the SetBatch.
+func BatchDelete[T any](sb *ygnmi.SetBatch, q ygnmi.SingletonQuery[T]) {
+	ygnmi.BatchDelete[T](sb, &singletonAsConfig[T]{SingletonQuery: q})
+}
+
+type singletonAsConfig[T any] struct {
+	ygnmi.SingletonQuery[T]
+}
```
wenovus, line 105: add a comment for its purpose please.

gnmi/client/client.go
```diff
@@ -69,3 +69,39 @@ func NewYGNMIClient() (*ygnmi.Client, error) {
 	}
 	return ygnmi.NewClient(gClient, ygnmi.WithTarget(viper.GetString("target")))
 }
+
```
wenovus, line 72: Perhaps rename this package from "client" to "gnmiclient" to be less generic?

dataplane/handlers/interface.go
```diff
@@ -224,26 +224,40 @@ func (ni *Interface) handleLinkUpdate(lu *netlink.LinkUpdate) {
...
+		iface.AdminStatus = oc.Interface_AdminStatus_UP
+	}
+	// TODO: handle other states.
 	var operStatus oc.E_Interface_OperStatus
-	switch lu.Attrs().OperState { // TODO: handle other states.
+	switch lu.Attrs().OperState {
 	case netlink.OperDown:
 		operStatus = oc.Interface_OperStatus_DOWN
-	case netlink.OperUp:
+	case netlink.OperUp, netlink.OperUnknown: // TAP interfaces are in unknown state until there is socket bound to them.
```
wenovus, line 237: where does the socket get attached? Is there just two ends to the TAP interface?

dataplane/handlers/interface.go
```diff
@@ -254,16 +268,27 @@ func (ni *Interface) handleAddrUpdate(au *netlink.AddrUpdate) {
...
 		}
 	} else {
 		if isV4 {
 			sub.GetOrCreateIpv4().DeleteAddress(ip)
+			client.BatchDelete(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv4().Address(ip).Ip().State())
+			client.BatchDelete(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv4().Address(ip).PrefixLength().State())
 		} else {
 			sub.GetOrCreateIpv6().DeleteAddress(ip)
+			client.BatchDelete(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv6().Address(ip).Ip().State())
+			client.BatchDelete(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv6().Address(ip).PrefixLength().State())
```
wenovus, line 286: nit: can just delete at Address(ip)

wenovus: Where is the Config() being set for these paths?

DanG100: done. 
config is set by the user through gnni.Set

dataplane/handlers/interface.go
DanG100: by some other process,

dataplane/handlers/interface.go
```diff
@@ -224,26 +224,40 @@ func (ni *Interface) handleLinkUpdate(lu *netlink.LinkUpdate) {
...
+		iface.AdminStatus = oc.Interface_AdminStatus_UP
+	}
+	// TODO: handle other states.
 	var operStatus oc.E_Interface_OperStatus
-	switch lu.Attrs().OperState { // TODO: handle other states.
+	switch lu.Attrs().OperState {
 	case netlink.OperDown:
 		operStatus = oc.Interface_OperStatus_DOWN
-	case netlink.OperUp:
+	case netlink.OperUp, netlink.OperUnknown: // TAP interface may be unknown state because the dataplane doesn't bind to its fd, so treat unknown as up.
```
wenovus, line 237: Interesting, so I guess the fact that the interface gets bound during port creation (via Build()) isn't at the right time, or isn't enough?

dataplane/handlers/interface.go
wenovus: I see, we're operating under the assumption that users will configure all interface paths before the test begins.

wenovus (approved):

dataplane/handlers/interface.go
DanG100: how else would the IPs be configured?

dataplane/handlers/interface.go
DanG100: i'm not sure what exactly cause the interface to be UP, but it works as long as its not down

PR #46, opened 2022-09-13
Allow the datastore to update state values by calling GnmiUpdate directly.

wenovus:
Also fix testagentlocal which was continuously updating the applied state even when there were no updates. Alas, easy to get things wrong without a test.

### DIFF
--- a/gnmi/gnmit/datastore.go
+++ b/gnmi/gnmit/datastore.go
@@ -27,6 +27,10 @@ func NewDatastoreServer(gnmiServer *GNMIServer) *DatastoreServer {
 }
 
 func (d *DatastoreServer) Set(_ context.Context, req *gpb.SetRequest) (*gpb.SetResponse, error) {
+	// TODO(wenbli): Reject values that modify config values. We only allow modifying state through this server.
+	// TODO(wenbli): Unmarshal to a struct without PreferShadowPath in
+	// order to validate that state paths are valid according to the
+	// schema.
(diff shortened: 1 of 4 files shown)


gnmi/gnmit/datastore.go
```diff
@@ -25,8 +27,26 @@ func NewDatastoreServer(gnmiServer *GNMIServer) *DatastoreServer {
 }
 
 func (d *DatastoreServer) Set(_ context.Context, req *gpb.SetRequest) (*gpb.SetResponse, error) {
-	if err := d.gnmiServer.set(req); err != nil {
+	if err := d.gnmiServer.set(req, false); err != nil {
```
DanG100, line 30: should update cache if the setrequest contains config?

wenovus: Thanks for pointing this out, I see two problems here:
1. the datastore should only process state updates, whereas the main gNMI server should only process config updates. Added TODO to enforce this rule since the solution should be similar.
2. We should still verify the validity of the received state values, but currently we ignore state values when unmarshalling due to `PreferShadowPath` so we aren't checking some of the state values being received. I will add another `set()` routine to process these state updates.

DanG100: sgtm with TODO

DanG100 (approved):

gnmi/gnmit/datastore.go
wenovus: thanks.

PR #47, opened 2022-09-14
TE2.1 subtest 1

wenovus:
I tested this a while back prior to the dataplane was added.

### DIFF
--- a/integration_tests/integration_test.go
+++ b/integration_tests/integration_test.go
@@ -1,0 +1,94 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


wenovus:
:facepalm:

wenovus:
Looks like shellcheck doesn't like comments as part of the ignore directive.

integration_tests/integration_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// This check assumes that 192.168.0.1 is reachable.
+	chk.HasResult(t, c.Results(t),
+		fluent.OperationResult().
+			WithIPv4Operation("1.1.1.1/30").
+			WithOperationType(constants.Add).
+			WithProgrammingResult(fluent.InstalledInRIB).
+			AsResult(),
+		chk.IgnoreOperationID(),
+	)
```
DanG100, line 93: is the plan to add traffic to this test?

wenovus: Yes this is essentially a sample of how gRIBI entries are installed. This PR doesn't need to be submitted since it's not a full test.

PR #50, opened 2022-09-15
Sysrib interface handlers and bug fix to datastore server 

wenovus:


sysrib/server.go
```diff
@@ -88,18 +95,87 @@ func (d *Dataplane) ProgramRoute(r *ResolvedRoute) error {
...
+}
+
+// monitorConnectedIntfs starts a gothread to check for connected prefixes from
+// connected interfaces and adds them to the sysrib. It returns an error if
+// there is an error before monitoring can begin.
+//
+// - gnmiServerAddr is the address of the central gNMI datastore.
+// - target is the name of the gNMI target.
+func (s *Server) monitorConnectedIntfs(gnmiServerAddr, target string) error {
+	var opts []grpc.DialOption
```
DanG100, line 123: maybe refactor this to gnmiclient package? 
gnmi.NewLocal()
gnmi.NewLocalWthSet()

sysrib/server.go
```diff
@@ -88,18 +95,87 @@ func (d *Dataplane) ProgramRoute(r *ResolvedRoute) error {
...
+	opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
DanG100, line 124: this won't work if tls is enabled

sysrib/server.go
```diff
@@ -88,18 +95,87 @@ func (d *Dataplane) ProgramRoute(r *ResolvedRoute) error {
...
+
+	b := &ocpath.Batch{}
+	b.AddPaths(
+		ocpath.Root().InterfaceAny().State().PathStruct(),
+	)
+
+	// TODO(wenbli): Ideally, this is implemented by watching more fine-grained paths.
+	// TODO(wenbli): Support interface removal.
+	go func() {
+		interfaceWatcher := ygnmi.Watch(
```
DanG100, line 145: nit: setup the watcher outside the goroutine and call await inside it

sysrib/server_test.go
```diff
@@ -627,33 +681,102 @@ func TestServer(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
+			grpcServer := grpc.NewServer()
```
DanG100, line 684: why make a server for every test case? you create the server once and delete /interfaces/ at the start of the test case

sysrib/server_test.go
```diff
@@ -627,33 +681,102 @@ func TestServer(t *testing.T) {
...
 			if err != nil {
 				t.Fatal(err)
 			}
 
-			for _, intf := range tt.inInterfaces {
-				if err := s.addInterfacePrefix(intf.name, intf.ifindex, intf.prefix, intf.niName); err != nil {
-					t.Fatal(err)
+			clientCtx, cancel := context.WithCancel(context.Background())
+			var sendErr, recvErr error
+			go func(ctx context.Context) {
```
DanG100, line 705: why do this in gorouting if you blocking waiting to goroutine to finish right after?

sysrib/server.go
wenovus: I'll do that in the next PR.

sysrib/server.go
wenovus: I'll also fix this in the next PR as part of refactoring.

sysrib/server.go
wenovus: Done.

sysrib/server_test.go
wenovus: Good point.. created TODO, I still need to implement deleting interfaces. This would be a good test of this ability.

sysrib/server_test.go
wenovus: Removed.

sysrib/server_test.go
DanG100: i doubt deleting interface is something we something we ever need to care about. (when would a user delete an interface) if restart the sysrib between test cases it will need see the delete

DanG100 (approved):

sysrib/server_test.go
wenovus: Got it, updated TODO's comment for this.

PR #55, opened 2022-09-22
Add hostname test for gNMI and convert systembase task to ygnmi

wenovus:
I had to modify the node name in `kne/topo.pb.txt` because Ondatra has a bug where for dut.Telemetry() only it uses res.Name() as the target. I'll submit the ondatra fix separately.

Added a makefile for less up arrow hitting.


integration_tests/system_test.go
```diff
@@ -0,0 +1,125 @@
...
+		{"Single Character", "x"},
+		{"Periods", "test.name.example"},
+		{"63 Characters", "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"},
+	}
+
+	dut := ondatra.DUT(t, "dut")
+
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			config := dut.Config().System().Hostname()
```
DanG100, line 47: here and below: use the new telemetry library. We want to delete the old one soon https://github.com/openconfig/ondatra/blob/main/gnmi/gnmi.go

DanG100:
> because Ondatra has a bug where for dut.Telemetry() only it uses res.Name() as the target. I'll submit the ondatra fix separately.

the dut.Telemetry() call is correct, the name of the device in topo.pb.txt should "lemming" and the target should be "lemming". The ID is the name from the testbed. I believe https://github.com/openconfig/ondatra/blob/main/gnmi/gnmi.go inverts that which needs to be fixed

integration_tests/system_test.go
wenovus: This library doesn't seem to have an API to get config paths.

DanG100: the library is ygnmi. gnmi.Get[string](t, ocpath.Root().System().Hostname().Config())

wenovus: Ah so it can't infer T correctly since the generic type is an interface. Perhaps ygnmi should have a GetConfig to help with this?

wenovus:
> > because Ondatra has a bug where for dut.Telemetry() only it uses res.Name() as the target. I'll submit the ondatra fix separately.
> 
> the dut.Telemetry() call is correct, the name of the device in topo.pb.txt should "lemming" and the target should be "lemming". The ID is the name from the testbed file. I believe https://github.com/openconfig/ondatra/blob/main/gnmi/gnmi.go inverts that which needs to be fixed

I'm still getting this error, perhaps the latest ondatra will fix it?

wenovus:
> > > because Ondatra has a bug where for dut.Telemetry() only it uses res.Name() as the target. I'll submit the ondatra fix separately.
> > 
> > 
> > the dut.Telemetry() call is correct, the name of the device in topo.pb.txt should "lemming" and the target should be "lemming". The ID is the name from the testbed file. I believe https://github.com/openconfig/ondatra/blob/main/gnmi/gnmi.go inverts that which needs to be fixed
> 
> I'm still getting this error, perhaps the latest ondatra will fix it?

NVM it's working now.

integration_tests/system_test.go
```diff
@@ -0,0 +1,127 @@
...
+	dut := ondatra.DUT(t, "dut")
+
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			config := ocpath.Root().System().Hostname().Config()
+			state := ocpath.Root().System().Hostname().State()
+
+			gnmi.Replace(t, dut, config, testCase.hostname)
+
+			t.Run("Get Hostname Config", func(t *testing.T) {
```
DanG100, line 54: why is this a subtest?

integration_tests/system_test.go
```diff
@@ -0,0 +1,127 @@
...
+	dut := ondatra.DUT(t, "dut")
+
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			config := ocpath.Root().System().DomainName().Config()
+			state := ocpath.Root().System().DomainName().State()
+
+			gnmi.Replace(t, dut, config, testCase.domainname)
+
+			t.Run("Get Domainname Config", func(t *testing.T) {
```
DanG100, line 105: same here

integration_tests/system_test.go
wenovus: I copied this from featureprofiles. This allows for each of these to have a separate PASS/FAIL message on the test output.

integration_tests/system_test.go
DanG100: we don't because if the type inference get fixed it will easier to remove all the newly unnecessary types than go and change all the instances of GetConfig back Get

integration_tests/system_test.go
```diff
@@ -0,0 +1,127 @@
...
+			gnmi.Replace(t, dut, config, testCase.hostname)
+
+			t.Run("Get Hostname Config", func(t *testing.T) {
+				configGot := gnmi.Get[string](t, dut, config)
+				if configGot != testCase.hostname {
+					t.Errorf("Config hostname: got %s, want %s", configGot, testCase.hostname)
+				}
+			})
+
+			t.Run("Get Hostname Telemetry", func(t *testing.T) {
```
DanG100, line 61: nit: i prefer "State" over "telemetry"

DanG100 (approved):

integration_tests/system_test.go
DanG100: sgtm

PR #56, opened 2022-09-26
Add layer3 punting

DanG100:


dataplane/forwarding/fwdport/ports/tap.go
```diff
@@ -0,0 +1,167 @@
...
+		return fmt.Errorf("ports: output actions for port %v failed, err %v", p, err)
+	}
+	return nil
+}
+
+func (p *tapPort) process() {
+	go func() {
+		buf := make([]byte, 1500) // TODO: MTU
+		for {
+			n, err := p.file.Read(buf)
```
wenovus, line 89: I think we should add a comment to explain why we made this implementation choice.

dataplane/forwarding/fwdport/ports/tap.go
```diff
@@ -0,0 +1,167 @@
...
+
+// Update updates the actions of the port.
+func (p *tapPort) Update(upd *fwdpb.PortUpdateDesc) error {
+	var err error
+	defer func() {
+		if err != nil {
+			p.Cleanup()
+		}
+	}()
+	kernelUpd, ok := upd.Port.(*fwdpb.PortUpdateDesc_Kernel)
```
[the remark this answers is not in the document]
DanG100: i added a todo for this. i want to think if we want a common type here. otherwise this would be the 3 instance of input and output proto message

dataplane/forwarding/fwdport/ports/tap.go
DanG100: added comment on the tabPort type

dataplane/handlers/interface.go
```diff
@@ -291,15 +288,21 @@ func (ni *Interface) handleAddrUpdate(ctx context.Context, au *netlink.AddrUpdat
...
 			gnmiclient.BatchUpdate(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv4().Address(ip).PrefixLength().State(), uint8(pl))
 		} else {
+			ipBytes = au.LinkAddress.IP.To16()
 			sub.GetOrCreateIpv6().GetOrCreateAddress(ip).PrefixLength = ygot.Uint8(uint8(pl))
 			gnmiclient.BatchUpdate(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv6().Address(ip).Ip().State(), au.LinkAddress.IP.String())
 			gnmiclient.BatchUpdate(sb, ocpath.Root().Interface(modelName).Subinterface(0).Ipv6().Address(ip).PrefixLength().State(), uint8(pl))
 		}
+		if err := engine.AddLayer3PuntRule(ctx, ni.fwd, modelName, ipBytes); err != nil {
+			log.Warningf("failed to add layer3 punt rule: %v", err)
+		}
```
wenovus, line 305: Forward all packets destined to this interface to the corresponding TAP interface.

dataplane/handlers/interface.go
```diff
@@ -329,7 +332,7 @@ func (ni *Interface) handleNeighborUpdate(ctx context.Context, nu *netlink.Neigh
 	sub := ni.getOrCreateInterface(modelName).GetOrCreateSubinterface(0)
 
 	if nu.Type == unix.RTM_DELNEIGH {
```
wenovus, line 334: switch statement?

wenovus (approved):

dataplane/handlers/interface.go
DanG100: aded

PR #57, opened 2022-09-26
Basic BGP Establish Test

wenovus:


bgp/bgp.go
```diff
@@ -0,0 +1,106 @@
+package main
```
wenovus, line 1: TODO: Delete this experimental file, it's not used by the test.

kne/twodut_config.yaml
```diff
@@ -0,0 +1,5 @@
+topology: /usr/local/google/home/wenbli/gocode/src/github.com/openconfig/lemming/kne/twodut_topo.pb.txt
```
DanG100, line 1: don't check in this file

kne/deploytwodut.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+cd "$(dirname "$0")"
+
+kne delete ./twodut_topo.pb.txt || true
+cd ..
+DOCKER_BUILDKIT=1 docker build . -t "lemming:latest"
+kind load docker-image lemming:latest --name kne
+cd -
+kne create ./twodut_topo.pb.txt
```
DanG100, line 10: ideally modify deploy.sh to have a topo argument. i can fix that in follow up PR

dataplane/handlers/interface.go
```diff
@@ -302,6 +302,8 @@ func (ni *Interface) handleAddrUpdate(ctx context.Context, au *netlink.AddrUpdat
 		}
 		if err := engine.AddLayer3PuntRule(ctx, ni.fwd, modelName, ipBytes); err != nil {
 			log.Warningf("failed to add layer3 punt rule: %v", err)
+		} else {
```
DanG100, line 305: remove this, i added logging in the other PR

DanG100 (approved):

kne/deploytwodut.sh
wenovus: It's a bit annoying because of the `dirname`, I'll let you decide what to do there.

PR #59, opened 2022-09-29
Add reconciler interface

DanG100:


gnmi/reconciler/reconciler.go
```diff
@@ -0,0 +1,40 @@
...
+	ID() string
+	// Start start the reconciliation loop.
+	// The client and target are connected to the local gNMI cache.
+	// An error returned during Start will cause lemming to exit, so Start should return non-retriable errors.
+	Start(ctx context.Context, client gpb.GNMIClient, target string) error
+	// Stop stops the reconciliation loop.
+	Stop(context.Context) error
+	// Validate is called after a SetRequest is checked for schema compliance, but before data is written to the cache.
+	// Reconcilers can validate a SetRequest for semantic correctness, config that can never be reconciled.
+	Validate(gpb.SetRequest) error
```
wenovus, line 37: We can pass this the root GoStruct instead so that it seems the whole view of the config.

gnmi/reconciler/reconciler.go
```diff
@@ -0,0 +1,40 @@
...
+	// ValidationPaths returns the set of path prefixes that the reconciler can validate.
+	ValidationPaths() []*gpb.Path
```
wenovus, line 39: What is the purpose of this?

DanG100: The Validate func is only called if the SetRequest contains paths which match the ValidationPaths.

wenovus: Would returning PathStructs be more user-friendly? e.g. wildcards
Or we should at least mention that the paths are query paths and thus can be wildcarded.

DanG100: sgtm

wenovus (approved):

PR #60, opened 2022-09-29
Modify dataplane to use local gnmi

DanG100:


gnmi/gnmiclient/cache.go
```diff
@@ -0,0 +1,125 @@
...
+// Subscribe implements gNMI Subscribe, by calling a gNMI server directly.
+func (cc *cacheClient) Subscribe(ctx context.Context, opts ...grpc.CallOption) (gpb.GNMI_SubscribeClient, error) {
+	errCh := make(chan error)
+	respCh := make(chan *gpb.SubscribeResponse, 10)
+	reqCh := make(chan *gpb.SubscribeRequest)
+
+	sub := &subServer{
+		respCh: respCh,
+		reqCh:  reqCh,
+		ctx:    peer.NewContext(ctx, &peer.Peer{}),
```
wenovus, line 58: What's the purpose of this?

gnmi/gnmiclient/cache.go
```diff
@@ -0,0 +1,125 @@
...
+	"io"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/peer"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+// NewCacheClient creates a gNMI client for the gNMI cache.
+// The client calls the server gRPC implementation directly to reduce overhead.
```
wenovus, line 28: The client calls the server gRPC implementation with a custom streaming gRPC implementation in order to bypass the regular gRPC wire marshalling/unmarshalling handling.

wenovus (approved):
Nice. The gnmi implementation and coalescing queue automatically handles pushback when `respCh` is filled, and so as long as we handle parallelism as you've done here, we can simply forward the responses to the client and skip the marshalling/unmarshalling that's done by the gRPC library. LGTM

gnmi/gnmiclient/cache.go
DanG100: the cache logs it, not setting it caused nil pointer

https://github.com/openconfig/lemming/blob/main/gnmi/subscribe/subscribe.go#L186

PR #63, opened 2022-09-30
Convert ygnmi GoBGP task

wenovus:
Also fixed the following:
* Fixed TODO for op state datastore service. It now behaves the same way as the config datastore.
* Removed the Task interface. The boilerplate savings are substantial.
* Imported latest version of ygot to simply datastore's unmarshalling code.

### DIFF
--- a/gnmi/gnmit/gnmit_test.go
+++ b/gnmi/gnmit/gnmit_test.go
@@ -149,7 +149,7 @@ func toUpd(r *gpb.SubscribeResponse) []*upd {
 // TestONCE tests the subscribe mode of gnmit.
 func TestONCE(t *testing.T) {
 	ctx := context.Background()
-	c, addr, err := New(ctx, nil, "localhost:0", "local", false, nil)
+	c, addr, err := New(ctx, nil, "localhost:0", "local", false)
 	if err != nil {
 		t.Fatalf("cannot start server, got err: %v", err)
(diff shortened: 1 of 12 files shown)


gnmi/gnmit/gnmit_test.go
```diff
@@ -149,7 +149,7 @@ func toUpd(r *gpb.SubscribeResponse) []*upd {
 // TestONCE tests the subscribe mode of gnmit.
 func TestONCE(t *testing.T) {
 	ctx := context.Background()
-	c, addr, err := New(ctx, nil, "localhost:0", "local", false, nil)
+	c, addr, err := New(ctx, nil, "localhost:0", "local", false)
```
DanG100, line 152: these tests should stop the server at end, i think that cause the test to fail

DanG100:
thanks for all this refactoring, LGTM pending test fix

gnmi/gnmit/gnmit_test.go
wenovus: This hit some rare race condition, essentially we can't have listeners listening to it before we try to delete the socket file. I'll take a look at fixing this in a follow-up.

DanG100: sgtm

DanG100 (approved):

PR #66, opened 2022-10-04
Unify gNMI datastore servers using context metadata

wenovus:
Considered using Peer but it cannot carry arbitrary information: https://pkg.go.dev/google.golang.org/grpc/peer

Still figuring how exactly to restructure gnmit/gnmi packages.


gnmi/gnmit/gnmit.go
```diff
@@ -366,20 +346,53 @@ func set(schema *ytypes.Schema, cache *cache.Cache, target string, req *gpb.SetR
 // Set is a prototype for a gNMI Set operation.
 // TODO(wenbli): Add unit test.
 func (s *GNMIServer) Set(ctx context.Context, req *gpb.SetRequest) (*gpb.SetResponse, error) {
-	s.c.smu.Lock()
-	defer s.c.smu.Unlock()
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
```
DanG100, line 350: we shouldn't require external users (ondatra tests) to have to set some metadata for set requests to work

gnmi/gnmiclient/client.go
```diff
@@ -21,56 +21,51 @@ import (
...
 
 // localClient is a client that connects to local gNMI cache.
 type localClient struct {
+	gnmiMode gnmistore.GNMIMode
 	gpb.GNMIClient
-	setClient gpb.GNMIClient
 }
 
 // Set uses the datastore client for Set, instead of the public cache endpoint.
 func (m *localClient) Set(ctx context.Context, in *gpb.SetRequest, opts ...grpc.CallOption) (*gpb.SetResponse, error) {
```
DanG100, line 39: move this to cacheclient and get rid of localClient entirely?

gnmi/gnmit/gnmit.go
wenovus: It by default uses ConfigMode, added comment to make that clearer.

DanG100: this still requires metadata to be set, which is always true

wenovus: Done.

gnmi/gnmiclient/client.go
wenovus: Posted.

gnmi/gnmistore/datastore.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package gnmistore contains a gNMI cache implementation that can be used as
+// an on-device/fake device implementation.
+package gnmistore
```
DanG100, line 17: optional: does this need to be in its own package?

DanG100 (approved):

gnmi/gnmistore/datastore.go
wenovus: I created this because there was a circular dependency. I'm looking to remove either gnmit/gnimstore once the server is cleaned up.

PR #68, opened 2022-10-04
Only use CacheClient as gNMI client helper.

wenovus:
Refactored existing code to use the same pattern as the dataplane for using this helper.

### DIFF
--- a/gnmi/gnmiclient/cache.go
+++ b/gnmi/gnmiclient/cache.go
@@ -16,13 +16,9 @@ package gnmiclient
 
 import (
 	"context"
-	"crypto/tls"
-	"fmt"
 	"io"
 
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
-	"google.golang.org/grpc/credentials/local"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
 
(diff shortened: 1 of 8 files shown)


gnmi/gnmiclient/cache.go
```diff
@@ -16,35 +16,75 @@ package gnmiclient
...
-// NewCacheClient creates a gNMI client for the gNMI cache.
+// New creates a state-based gNMI client for the gNMI cache.
 // The client calls the server gRPC implementation with a custom streaming gRPC implementation
 // in order to bypass the regular gRPC wire marshalling/unmarshalling handling.
-// TODO: refactor gNMI so that a single server can be used here.
-func NewCacheClient(srv gpb.GNMIServer, setClient gpb.GNMIClient) gpb.GNMIClient {
-	return &cacheClient{
-		srv:       srv,
-		setClient: setClient,
+func New(srv gpb.GNMIServer, port int, enableTLS bool) (gpb.GNMIClient, error) {
```
DanG100, line 36: don't create a client, just call the server Set directly?

gnmi/gnmiclient/cache.go
```diff
@@ -19,36 +19,47 @@ import (
...
-	}
+		gnmiMode: gnmistore.StateMode,
+		srv:      srv,
+	}, nil
+}
+
+// NewReadOnly creates a config-based gNMI client for the gNMI cache.
+// The client calls the server gRPC implementation with a custom streaming gRPC implementation
+// in order to bypass the regular gRPC wire marshalling/unmarshalling handling.
+func NewReadOnly(srv gpb.GNMIServer) (gpb.GNMIClient, error) {
```
DanG100, line 42: do we need this? i don't see where it's used

DanG100 (approved):

gnmi/gnmiclient/cache.go
wenovus: True, I don't see this being used.

PR #72, opened 2022-10-05
Migrate tasks to reconcilers

DanG100:


gnmi/gnmi.go
```diff
@@ -277,7 +292,7 @@ func (s *Server) Set(ctx context.Context, req *gpb.SetRequest) (*gpb.SetResponse
 
 		// TODO(wenbli): Reject paths that try to modify read-only values.
 		// TODO(wenbli): Question: what to do if there are operational-state values in a container that is specified to be replaced or deleted?
-		err := set(s.configSchema, s.c.cache, s.c.name, req, true)
+		err := set(s.configSchema, s.c.cache, s.c.name, req, true, s.validators)
 
 		// TODO(wenbli): Currently the SetResponse is not filled.
 		return &gpb.SetResponse{
```
wenovus, line 298: It looks like the status.Error() is missing here.

gnmi/gnmi.go
```diff
@@ -215,10 +225,10 @@ func updateCacheNotifs(cache *cache.Cache, nos []*gpb.Notification, target, orig
 //
 // update indicates whether to update the cache with the values from the set
 // request.
-func set(schema *ytypes.Schema, cache *cache.Cache, target string, req *gpb.SetRequest, preferShadowPath bool) error {
+func set(schema *ytypes.Schema, cache *cache.Cache, target string, req *gpb.SetRequest, preferShadowPath bool, validators []func(*oc.Root) error) error {
 	prevRoot, err := ygot.DeepCopy(schema.Root)
 	if err != nil {
-		return fmt.Errorf("gnmi: failed to ygot.DeepCopy the cached root object: %v", err)
+		return status.Errorf(codes.Internal, "failed to ygot.DeepCopy the cached root object: %v", err)
```
wenovus, line 231: This is an internal function, should it rather return a normal error, and leave the task of wrapping it in a gRPC status to the server implementation? (i.e. Set in this case)

gnmi/gnmi.go
DanG100: everything returned by set is status.Error so we don't want to wrap it twice

gnmi/gnmi.go
DanG100: this returns different codes so we can't wrap it universally

wenovus: OK, can you add a comment on this behaviour for set?

gnmi/gnmi.go
```diff
@@ -292,7 +307,7 @@ func (s *Server) Set(ctx context.Context, req *gpb.SetRequest) (*gpb.SetResponse
 			return s.UnimplementedGNMIServer.Set(ctx, req)
 		}
 		// TODO(wenbli): Reject values that modify config values. We only allow modifying state in this mode.
-		if err := set(s.stateSchema, s.c.cache, s.c.name, req, false); err != nil {
+		if err := set(s.stateSchema, s.c.cache, s.c.name, req, false, nil); err != nil {
 			return &gpb.SetResponse{}, status.Errorf(codes.Aborted, "%v", err)
```
wenovus, line 311: Remove the wrap here.

gnmi/gnmiclient/cache.go
```diff
@@ -23,29 +23,43 @@ import (
...
+const (
+	// GNMIModeMetadataKey is the context metadata key used to specify the
+	// mode in which the gNMI server should operate.
+	GNMIModeMetadataKey = "gnmi-mode"
+	// ConfigMode indicates that the gNMI service will allow updates to
+	// intended configuration, but not operational state values.
+	ConfigMode Mode = "config"
+	// StateMode indicates that the gNMI service will allow updates to
+	// operational state, but not intended configuration values.
+	StateMode Mode = "state"
```
wenovus, line 40: Should these be owned by the client package or the server package? Move this back to the new package gnmistore to avoid the circular reference?

gnmi/gnmi.go
```diff
@@ -338,3 +326,24 @@ func (s *Server) Get(ctx context.Context, req *gpb.GetRequest) (*gpb.GetResponse
 		return nil, status.Errorf(codes.DataLoss, "Unknown message type: %T", resp)
 	}
 }
+
+// StartReconcilers starts all the reconcilers.
+func (s *Server) StartReconcilers(ctx context.Context) error {
+	c := gnmiclient.New(s)
```
wenovus, line 332: Should this be an input to avoid importing the client package?

gnmi/gnmiclient/cache.go
DanG100: moved to gnmi package

gnmi/gnmi.go
DanG100: moved to gnmi package

wenovus (approved):

PR #78, opened 2022-10-11
Query and Update Packet Counters (in-pkt/out-pkt)

wenovus:
Currently this is done by polling the forwarding service once per second. Ideally the forwarding service supports streaming the counter values, but I think this is good enough for now. Added a TODO in the forwarding service definition.

For some reason I had to downgrade Ondatra for the traffic test to work (I think the error was target "lemming" not found) -- Could it be due to https://github.com/openconfig/ondatra/commit/3338c3259dd8419b059443cadd238841a2472015#diff-14271c87af5568921ff5c49b376f4d9b0233c11ab49660ec5f6bfd0a1f7174c4 ?

### DIFF
--- a/dataplane/handlers/interface.go
+++ b/dataplane/handlers/interface.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/openconfig/lemming/dataplane/internal/engine"
 	"github.com/openconfig/lemming/dataplane/internal/kernel"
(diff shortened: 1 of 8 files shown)


dataplane/handlers/interface.go
```diff
@@ -132,6 +133,52 @@ func (ni *Interface) Start(ctx context.Context) error {
...
+						if _, err := gnmiclient.Replace(ctx, ni.c, ocpath.Root().Interface(intfName).Counters().InPkts().State(), counter.Value); err != nil {
+							log.Errorf("interface handler: %v", err)
+						}
+					case fwdpb.CounterId_COUNTER_ID_TX_PACKETS:
+						if _, err := gnmiclient.Replace(ctx, ni.c, ocpath.Root().Interface(intfName).Counters().OutPkts().State(), counter.Value); err != nil {
+							log.Errorf("interface handler: %v", err)
+						}
+					}
+				}
+			}
```
DanG100, line 178: move func to method on the handler
func (ni *Interface) updateCounterrs() {}

dataplane/handlers/interface.go
```diff
@@ -132,6 +133,52 @@ func (ni *Interface) Start(ctx context.Context) error {
 		}
 	}()
 
+	// Gothread for updating counters.
+	tick := time.NewTicker(time.Second)
```
DanG100, line 137: stop in the ticker in Stop()

dataplane/handlers/interface.go
```diff
@@ -132,6 +133,52 @@ func (ni *Interface) Start(ctx context.Context) error {
 		}
 	}()
 
+	// Gothread for updating counters.
```
DanG100, line 136: Gothread -> goroutine

DanG100 (approved):

PR #79, opened 2022-10-11
Use context metadata to allow state updates to specify SetRequest timestamp.

wenovus:
Idea from @DanG100

### DIFF
--- a/gnmi/cache.go
+++ b/gnmi/cache.go
@@ -46,6 +46,14 @@ const (
 	TimestampMetadataKey = "gnmi-timestamp"
 )
 
+// appendToIncomingContext returns a new context with the provided kv merged
+// with any existing metadata in the context. Please refer to the documentation
+// of Pairs for a description of kv.
+func appendToIncomingContext(ctx context.Context, kv ...string) context.Context {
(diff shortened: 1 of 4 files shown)


gnmi/cache.go
```diff
@@ -38,8 +39,25 @@ const (
...
+// AddTimestampMetadata adds a gNMI timestamp metadata to the context.
+//
+// - ctx is the context to be used for accessing lemming's internal datastore.
+// - timestamp is the number of nanoseconds since Epoch.
+//
+// NOTE: The output of this function should only be used to call into the
+// internal lemming gNMI server. This is because it adds an incoming rather
+// than an outgoing context metadata to skip regular protobuf handling.
+func AddTimestampMetadata(ctx context.Context, timestamp int64) context.Context {
+	return metadata.NewIncomingContext(ctx, metadata.Pairs(TimestampMetadataKey, strconv.FormatInt(timestamp, 10)))
```
DanG100, line 58: this will overwrite the metadata is the context already has metadata.
instead, get the metadata if it exists and update it or create a new one

wenovus: Good catch, added handling.

DanG100 (approved):

PR #88, opened 2022-10-27
Basic presubmit

DanG100:


cloudbuild/presubmit.sh
```diff
@@ -0,0 +1,13 @@
...
+
+export PATH=${PATH}:/usr/local/go/bin
+gopath=$(go env GOPATH)
+export PATH=${PATH}:$gopath/bin
+
+cd /tmp/workspace
+kne deploy ~/kne-internal/deploy/kne/kind-bridge.yaml
+make deploy itest
+make clean
+make deploy2 itest2
```
wenovus, line 13: Does it make sense to split these commands in different build steps so that it's easier to know which step or test failed? I'm not familiar with all the env args to know whether you can simply create another build step for it.

wenovus: Interesting, where is the VM being created? so with a private pool it's not possible to split up the steps?

DanG100: yaml file controls the workflow, in this case a single step vm creation and running the script

wenovus: So the environment variables are creating the VM and a second step would not be able to access the VM? Is there some documentation for it? Perhaps add a README here?

DanG100: the step is create the vm and the run the script

wenovus (approved):

PR #92, opened 2022-11-01
`extra_tail_actions` for SetRoute in dataplane API.

wenovus:
Also fix the go package paths.

Implementing this is not as simple as just encapping since there is the MAC header that we need to deal with. I didn't see an action to insert a header to the packet.


proto/dataplane/dataplane.proto
```diff
@@ -53,6 +55,10 @@ message NextHop {
   string port = 1;
   string ip = 2;
   uint64 weight = 3;
+  // extra_tail_actions are extra actions that are applied just prior to
+  // transmitting the packet to the output port (at which time the output
+  // actions are applied).
+  repeated forwarding.ActionDesc extra_tail_actions = 4;
```
DanG100, line 61: what about  something like `pre_transmit_acts`? to make it more clear when those actions are run

DanG100 (approved):

proto/dataplane/dataplane.proto
wenovus: thanks, done.

PR #95, opened 2022-11-10
Fix lint style issues and enable linter dataplane

DanG100:


wenovus (approved):
Thanks for doing this!

PR #102, opened 2022-11-19
sysrib Encap rule programming for BGP GUE

wenovus:
Does not include the reconciler logic. Will be added in following PR.


sysrib/server.go
```diff
@@ -521,15 +589,18 @@ func convertZebraRoute(niName string, zroute *zebra.IPRouteBody) *Route {
 			NetworkInstance: niName,
 		})
 	}
+	var routePref RoutePreference
+	switch zroute.Type {
+	case zebra.RouteBGP:
+		routePref.AdminDistance = 20
```
DanG100, line 595: should this be configureable? if so TODO, if not make it const

sysrib/server.go
```diff
@@ -281,6 +294,54 @@ func prefixString(prefix *pb.Prefix) (string, error) {
...
+	}
+	udp := layers.UDP{
+		SrcPort: 0, // TODO(wenbli): Implement hashing for srcPort.
+		DstPort: layers.UDPPort(binary.BigEndian.Uint16(gueHeaders.dstPort[:])),
+		Length:  34,
+	}
+
+	buf := gopacket.NewSerializeBuffer()
+	opts := gopacket.SerializeOptions{}
+	if err := udp.SerializeTo(buf, opts); err != nil {
```
DanG100, line 321: can simplify to gopacket.SerializeLayers(buf, opts, ip, upd)

sysrib/sysrib.go
```diff
@@ -48,9 +54,71 @@ type NIRIB struct {
...
+	}
+	ok, policy := sr.GUEPolicies.FindDeepestTag(*addr)
+	if !ok {
+		return GUEHeaders{}, false, nil
+	}
+	ip := net.ParseIP(address)
+	if ip == nil {
+		return GUEHeaders{}, false, fmt.Errorf("cannot parse IP address: %q", address)
+	}
+	switch policy.isV6 {
```
DanG100, line 96: change to if/else

sysrib/server.go
```diff
@@ -281,6 +294,54 @@ func prefixString(prefix *pb.Prefix) (string, error) {
...
+func gueActions(gueHeaders GUEHeaders) ([]*fwdpb.ActionDesc, error) {
+	ip := layers.IPv4{
+		Version:  4,
+		IHL:      5,
+		Length:   74,
+		Protocol: layers.IPProtocolUDP,
+		SrcIP:    gueHeaders.srcIP4[:],
+		DstIP:    gueHeaders.dstIP4[:],
+	}
+	if gueHeaders.isV6 {
```
DanG100, line 308: do you need to make layers.IPv6 header instead?

sysrib/sysrib.go
```diff
@@ -48,9 +54,71 @@ type NIRIB struct {
...
 }
 
+// GUEPolicy represents the static values in the IP and UDP headers that are
+// to encapsulate the packet.
+//
+// srcPort is a hash.
+// dstIP is the nexthop of the BGP route.
+type GUEPolicy struct {
+	dstPort [2]byte
+	srcIP4  [4]byte
```
DanG100, line 64: i think this code would be simpler if you had one slice srcIP since you already is ivV6 bool

sysrib/sysrib.go
```diff
@@ -48,9 +54,71 @@ type NIRIB struct {
 	IPV4 *generics_tree.TreeV4[*Route]
 }
 
+// GUEPolicy represents the static values in the IP and UDP headers that are
+// to encapsulate the packet.
+//
+// srcPort is a hash.
+// dstIP is the nexthop of the BGP route.
+type GUEPolicy struct {
+	dstPort [2]byte
```
DanG100, line 63: uint16?

sysrib/server.go
wenovus: Made const.

sysrib/sysrib.go
wenovus: I wanted to use an array type here to keep the types tight.

sysrib/server.go
```diff
@@ -281,6 +301,51 @@ func prefixString(prefix *pb.Prefix) (string, error) {
...
+		}
+	}
+	udp := &layers.UDP{
+		SrcPort: 0, // TODO(wenbli): Implement hashing for srcPort.
+		DstPort: layers.UDPPort(gueHeaders.dstPort),
+		Length:  34,
+	}
+
+	buf := gopacket.NewSerializeBuffer()
+	gopacket.SerializeLayers(buf, gopacket.SerializeOptions{}, ip, udp)
```
DanG100, line 330: this func returns an error

sysrib/sysrib.go
DanG100: i'm not sure i understand i understand what advantage this has considering you check the boolean every time these are used. you also wouldn't convert to and from slice/array every time you use these

wenovus: Oh I remember now, this type has to be hashable in order to do the diffing of the resolved nexthops, so a slice, or net.IP, wouldn't work.

DanG100 (approved):

PR #105, opened 2022-11-28
Add route reconciler

DanG100:


dataplane/handlers/routes.go
```diff
@@ -0,0 +1,108 @@
...
+func NewRoute(fwd fwdpb.ServiceClient) *reconciler.BuiltReconciler {
+	r := &route{
+		fwd: fwd,
+	}
+	return reconciler.NewBuilder("dataplane-routes").WithStart(r.start).Build()
+}
+
+func (r *route) start(ctx context.Context, client *ygnmi.Client) error {
+	r.w = ygnmi.WatchAll(ctx, client, routesQuery, func(v *ygnmi.Value[*dpb.InsertRouteRequest]) error {
+		route, present := v.Val()
```
wenovus, line 62: Does present reset to false if a datapoint is completely deleted? It looks like the extract function will always return true, so even on a deletion, I'm suspecting it might return true for present. https://github.com/openconfig/ygnmi/blob/main/schemaless/schemaless.go#L89

Should we add a test for watchAll in ygnmi?

dataplane/handlers/routes.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		if !present {
+			if err := engine.DeleteIPRoute(ctx, r.fwd, isIPv4, ipNet.IP, ipNet.Mask); err != nil {
+				log.Warningf("failed to delete route: %v", err)
+				return ygnmi.Continue
+			}
+			return ygnmi.Continue
+		}
+		if err := engine.AddIPRoute(ctx, r.fwd, isIPv4, ip, ipNet.Mask, route.GetNextHops()); err != nil {
```
wenovus, line 93: Is this robust to adding a route with the same key?

dataplane/handlers/routes.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		ip := ipNet.IP.To4()
+		isIPv4 := true
+		if ip == nil {
+			ip = ipNet.IP.To16()
+			isIPv4 = false
+		}
+
+		if !present {
+			if err := engine.DeleteIPRoute(ctx, r.fwd, isIPv4, ipNet.IP, ipNet.Mask); err != nil {
```
wenovus, line 87: Is this robust to deleting a route with the same key?

gnmi/gnmi.go
```diff
@@ -256,6 +256,56 @@ func set(schema *ytypes.Schema, cache *cache.Cache, target string, req *gpb.SetR
...
+		}
+	}
+	if hasInternal {
+		if err := s.c.cache.GnmiUpdate(notif); err != nil {
+			return true, err
+		}
+	}
+
+	notif.Delete = nil
+
```
wenovus, line 289: update hasInternal to false here.

gnmi/gnmi_test.go
```diff
@@ -588,6 +588,116 @@ func TestSetState(t *testing.T) {
 	}
 }
 
+func TestSetInternal(t *testing.T) {
```
wenovus, line 591: // TestSetInternal tests that the server is able to handle schemaless queries.

dataplane/handlers/routes.go
DanG100: fixed in https://github.com/openconfig/ygnmi/pull/74

dataplane/handlers/routes.go
DanG100: yeah, its really and/update

dataplane/handlers/routes.go
DanG100: deleting something twice will return an error, which should never happen

wenovus (approved):

PR #107, opened 2022-11-30
Change sysrib to use gNMI cache

DanG100:
I didn't change the dataplane interface thingy because the tests rely that. Could be changed in the future however


lemming.go
```diff
@@ -107,9 +102,13 @@ func New(lis net.Listener, targetName, zapiURL string, opts ...grpc.ServerOption
...
 
 	cacheClient := gnmiServer.LocalClient()
+	yclient, err := ygnmi.NewClient(cacheClient, ygnmi.WithTarget(targetName))
+	if err != nil {
+		return nil, err
+	}
 
 	log.Infof("starting sysrib")
-	sysribServer, err := sysrib.New(sysDataplane)
+	sysribServer, err := sysrib.New(&sysrib.Dataplane{Client: yclient})
```
wenovus, line 111: Remove this parameter and in sysrib/server.go, within func (s *Server) Start, set the ygnmi client for the dataplaneAPI field.

dataplane/server.go
```diff
@@ -126,57 +111,6 @@ func (d *Dataplane) Stop(ctx context.Context) error {
 	return nil
 }
 
-// InsertRoute inserts a route into the dataplane.
-func (d *Dataplane) InsertRoute(ctx context.Context, route *dpb.InsertRouteRequest) (*dpb.InsertRouteResponse, error) {
-	if len(route.NextHops) == 0 {
```
wenovus, line 131: Keep this check in dataplane/handlers/routes.go to avoid an index error when engine.AddIPRoute is called.

wenovus (approved):

PR #114, opened 2022-12-12
A few more operator fixes

DanG100:


Makefile
```diff
@@ -5,14 +5,14 @@ clean:
 
 .PHONY: load 
 load:
-	DOCKER_BUILDKIT=1 docker build . -f Dockerfile.lemming -t "lemming:latest"
-	kind load docker-image lemming:latest --name kne
+	DOCKER_BUILDKIT=1 docker build . -f Dockerfile.lemming -t "us-west1-docker.pkg.dev/openconfig-lemming/release/lemming:ga"
+	kind load docker-image us-west1-docker.pkg.dev/openconfig-lemming/release/lemming:ga --name kne
```
wenovus, line 9: when I'm developing how do I know I'm not overwriting someone else's image?

wenovus: Can you explain here why this full name is used?

DanG100: as of recently, kne will use `us-west1-docker.pkg.dev/openconfig-lemming/release/lemming:ga` as the default image for lemming. calling `make load` will build the image from src and use that one instead pulling the image remotely.

images names are treated as opaque strings not directories `us-west1-docker.pkg.dev/openconfig-lemming/release/lemming:ga` != `lemming:ga`. `lemming:ga` is an actually something`docker.io/lemming:ga` which also doesn't exist

wenovus: Thanks!

wenovus (approved):

PR #115, opened 2022-12-12
Add support for delete MAC and IP address

DanG100:


dataplane/handlers/interface.go
```diff
@@ -213,61 +213,102 @@ func (ni *Interface) reconcile(config *oc.Interface) {
...
-
-		if configIP != nil && configPL != nil && (stateIP == nil || *statePL != *configPL) {
-			log.V(1).Infof("Config IP: %v, Config PL: %v. State IP: %v, State PL: %v", addr.GetIp(), addr.GetPrefixLength(), stateAddr.GetIp(), stateAddr.GetPrefixLength())
-			log.V(2).Infof("setting interface %s ip %s/%d", tapName, *configIP, *configPL)
-			if err := kernel.SetInterfaceIP(tapName, *configIP, int(*configPL)); err != nil {
+
+	type addrPair struct {
+		cfgIP, stateIP *string
+		cfgPL, statePL *uint8
+	}
```
wenovus, line 239: prefixPair

dataplane/handlers/interface.go
```diff
@@ -213,61 +213,102 @@ func (ni *Interface) reconcile(config *oc.Interface) {
...
-				return
+			}
+		}
+	} else {
+		if state.GetEthernet().GetHwMacAddress() != state.GetEthernet().GetMacAddress() {
+			log.V(1).Infof("resetting interface %s hw-addr %q", tapName, state.GetEthernet().GetHwMacAddress())
+			if err := kernel.SetInterfaceHWAddr(config.GetName(), state.GetEthernet().GetHwMacAddress()); err != nil {
+				log.Warningf("Failed to set mac address of port: %v", err)
 			}
 		}
```
wenovus, line 224: Add a comment on why this gets reset (perhaps link to https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-ethernet-state-mac-address)

dataplane/handlers/interface.go
```diff
@@ -213,61 +213,102 @@ func (ni *Interface) reconcile(config *oc.Interface) {
...
+		if pairAddr := config.GetSubinterface(0).GetIpv6().GetAddress(addr.GetIp()); pairAddr != nil {
+			pair.cfgIP = pairAddr.Ip
+			pair.cfgPL = pairAddr.PrefixLength
+		}
+		maybeDeletedPairs = append(maybeDeletedPairs, pair)
+	}
+
+	for _, pair := range maybeDeletedPairs {
+		// If an IP exists in state, but not in config, remove the IP.
+		if pair.stateIP != nil && pair.statePL != nil && (pair.cfgIP == nil && pair.cfgPL == nil) {
```
wenovus, line 268: brackets around the state variables. Also I'm wondering if it should be `||` on the state and the config.

dataplane/handlers/interface.go
```diff
@@ -213,61 +213,102 @@ func (ni *Interface) reconcile(config *oc.Interface) {
...
+		pair := &addrPair{
+			cfgIP: addr.Ip,
+			cfgPL: addr.PrefixLength,
+		}
+		if pairAddr := state.GetSubinterface(0).GetIpv4().GetAddress(addr.GetIp()); pairAddr != nil {
+			pair.stateIP = pairAddr.Ip
+			pair.statePL = pairAddr.PrefixLength
+		}
+		maybeAddedPairs = append(maybeAddedPairs, pair)
+	}
```
wenovus, line 289: This looks to be a dup? Should just have one but the name is `allInterfacePairs` instead of `maybeAddedPairs`?

wenovus: As discussed this is not a dup. Would `allInterfacePairs` still be an improvement since you can merge the SetInterfaceIP/DeleteInterfaceIP bodies?

dataplane/handlers/interface.go
DanG100: i wanted to make sure that the config is completely gone or completely added before making any change

dataplane/handlers/interface.go
DanG100: done

dataplane/handlers/interface.go
DanG100: done

dataplane/handlers/interface.go
DanG100: done

dataplane/handlers/interface.go
```diff
@@ -207,67 +207,105 @@ func (ni *Interface) reconcile(config *oc.Interface) {
...
-			}
+		pair := &prefixPair{
+			cfgIP: addr.Ip,
+			cfgPL: addr.PrefixLength,
+		}
+		if pairAddr := state.GetSubinterface(0).GetIpv4().GetAddress(addr.GetIp()); pairAddr != nil {
+			pair.stateIP = pairAddr.Ip
+			pair.statePL = pairAddr.PrefixLength
 		}
+		interfacePairs = append(interfacePairs, pair)
```
wenovus, line 277: I guess for the second pair of for loops you should only add if the state doesn't exist, as otherwise you have a duplicate entry.

wenovus (approved):
Is there a simple way to sanity test this?

dataplane/handlers/interface.go
DanG100: we could have the same state IP but a different prefix length, so we need to add it. the duplicate entry shouldn't cause an issue

DanG100:
i need to refactor this a bit more before i can easily add unit tests. there's already a todo for this

PR #116, opened 2022-12-13
Fix BGP GUE to accommodate understanding that IPv4-mapped IPv6 packets can go over IPv4 routes.

wenovus:
Also,
- Increase timeout of BGP establishment to 120 seconds. There is some slowdown going on in the system that we may need to investigate soon.
- Change log level of gNMI messages to 2.


integration_tests/twodut_oneotg_tests/bgp_triggered_gue/bgp_triggered_gue_test.go
```diff
@@ -357,6 +358,7 @@ func testTrafficAndEncap(t *testing.T, ate *ondatra.ATEDevice, top gosnappi.Conf
 			nl := pkt.NetworkLayer()
 			if nl == nil {
 				t.Fatalf("packet doesn't have network layer: %s", pkt.Dump())
```
DanG100, line 360: Fatal and continue. either change to Errorf or remove continue

DanG100 (approved):

PR #119, opened 2023-01-04
Use 2 go.mods for operator and lemming

DanG100:
This lets us two independent versions for the operator and lemming without dealing with two repos

https://go.dev/doc/modules/managing-source


wenovus (approved):
I assume the operator is independent of the lemming image?

wenovus (approved):

PR #126, opened 2023-01-13
Basic support for IPv4 static routes

wenovus:


sysrib/static.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+	}
+}
+
+// monitorStaticRoutes starts a gothread to check for static route
+// configuration changes.
+// It returns an error if there is an error before monitoring can begin.
+func (s *Server) monitorStaticRoutes(yclient *ygnmi.Client) error {
+	b := &ocpath.Batch{}
+	staticroot := ocpath.Root().NetworkInstance(fakedevice.DefaultNetworkInstance).Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "DEFAULT")
```
DanG100, line 63: should this a const? configureable?

sysrib/static.go
```diff
@@ -0,0 +1,107 @@
...
+		context.Background(),
+		yclient,
+		b.Config(),
+		func(root *ygnmi.Value[*oc.Root]) error {
+			rootVal, ok := root.Val()
+			if !ok {
+				return ygnmi.Continue
+			}
+			staticp := rootVal.GetOrCreateNetworkInstance(fakedevice.DefaultNetworkInstance).GetOrCreateProtocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "DEFAULT")
+			if staticp == nil {
```
DanG100, line 84: if you're calling GetOrCreate above when would this be nil?

sysrib/static.go
wenovus: Done.

sysrib/static.go
wenovus: Removed.

DanG100 (approved):

PR #130, opened 2023-01-20
Sysrib server supports IPv6 static routes

wenovus:
Added IPv6 static route traffic test.


sysrib/server.go
```diff
@@ -622,6 +578,66 @@ func (s *Server) ResolveAndProgramDiff() error {
 	return nil
 }
 
+func (s *Server) resolveAndProgramDiffAux(niName string, ni *NIRIB, address string, newResolvedRoutes map[RouteKey]*Route) {
```
DanG100, line 581: is there a more descriptive name for this?

wenovus: It's a helper function and what it does is strange as a standalone function.

I've added comments to clarify its behaviour.

DanG100 (approved):

PR #132, opened 2023-01-25
Support IPv6 BGP-triggered GUE and add some debugging statements

wenovus:


wenovus:
@DanG100 Does the dataplane support configuring both IPv4 and IPv6 addresses? If so I can later look at merging the IPv4 and IPv6 tests into a single one.

bgp/debug.go
```diff
@@ -0,0 +1,17 @@
+package bgp
```
DanG100, line 1: license header

sysrib/server_test.go
```diff
@@ -156,6 +156,173 @@ func configureInterface(t *testing.T, intf *AddIntfAction, yclient *ygnmi.Client
 	}
 }
 
+var (
+	// v4v6Mapper maps IPv4 addresses to IPv6 addresses for testing v6.
+	v4v6Mapper = map[string]string{
```
DanG100, line 161: can this be func instead a hardcoded map?

sysrib/server_test.go
```diff
@@ -156,6 +156,173 @@ func configureInterface(t *testing.T, intf *AddIntfAction, yclient *ygnmi.Client
...
+				isV6:      false,
+			},
+			dstIP4: [4]byte{192, 168, 1, 42},
+		}: {
+			GUEPolicy: GUEPolicy{
+				dstPortv6: 8,
+				srcIP6:    [16]byte{42, 42, 42, 42, 42},
+				isV6:      true,
+			},
+			dstIP6: [16]byte{0x20, 0x1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x42},
```
DanG100, line 213: maybe a test utility to go from ip str to array might be nice

DanG100:
> @DanG100 Does the dataplane support configuring both IPv4 and IPv6 addresses and forwarding both traffic at once? If so I can later look at merging the IPv4 and IPv6 tests into a single one.

that should be supported

sysrib/server_test.go
wenovus: `mapAddress` uses this map to convert the addresses.

sysrib/server_test.go
wenovus: There is the tradeoff of [16]byte forcing a very precise definition, so it's not yet clear that such a utility will be better. Maybe with more use cases we will see a pattern.

sysrib/server_test.go
wenovus: Done.

DanG100 (approved):

PR #140, opened 2023-02-09
Initial impl of insert in gNSI pathz trie

DanG100:


gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,197 @@
...
+		// Normalize path string by pruning wildcard keys.
+		pathStr, err := elemToString(elem.Name, elem.Key)
+		if err != nil {
+			return fmt.Errorf("invalid path element: %v", err)
+		}
+
+		// The rank of the child is the number of non-wildcards key (if any).
+		rank := len(setKeys(elem.Key))
+		if rank >= len(node.childrenByRank) {
+			node.childrenByRank = append(node.childrenByRank, make([]map[string]*trieNode, rank-len(node.childrenByRank)+1)...)
```
wenovus, line 75: I can't understand this part of the code. Can you explain this line?

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,197 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// elemToString returns a formatted string representation of a single path elem.
+// wildcard keys are pruned from the resulting string
+func elemToString(name string, kv map[string]string) (string, error) {
```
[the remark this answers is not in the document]
DanG100: added todo

gnsi/acltrie/trie.go
wenovus: Suggest to make this a generic utility. e.g. s = slices.Grow(s, newLen-len(s))[:newLen]) -- example is inlined.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	node.hasPolicy = true
+
+	return nil
+}
+
+// checkPathConflict returns false if the candidate path overlaps with the accepted path.
+// The candidate path conflicts with the accepted path if its name matches and
+// all keys match (where * matches with anything).
+func checkPathConflict(accepted, candidate *gpb.PathElem) bool {
```
wenovus, line 139: A nit is `accepted, candidate` sounds as if there is a difference between them. Would `a, b` just as accurate? Does conflict detection depend on the order of the paths?

gnsi/acltrie/trie_test.go
```diff
@@ -0,0 +1,320 @@
...
+
+func mustPath(s string) *gpb.Path {
+	p, err := ygot.StringToStructuredPath(s)
+	if err != nil {
+		panic(fmt.Sprintf("cannot parse subscription path %s, %v", s, err))
+	}
+	return p
+}
+
+func TestInsert(t *testing.T) {
```
wenovus, line 36: I think a slightly better test would be to check the state of the whole trie. This only checks for errors but doesn't check whether the tree was constructed correctly.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	for k, v := range candidate.Key {
+		acceptedV, ok := accepted.Key[k]
+		if !ok || acceptedV == "*" { // Candidate key matches against wildcard accepted.
+			continue
+		}
+		if v == "*" { // Candidate is wildcard, matches against anything.
+			continue
+		}
```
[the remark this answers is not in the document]
DanG100: see test case "failure ambiguous list keys"

gnsi/acltrie/trie_test.go
DanG100: if we do that the test become very implementation specific. i'll implement Get if the next PR which will get retrieving policies

gnsi/acltrie/trie.go
DanG100: slice.Grow only extends the capacity not the length of the slice

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,200 @@
...
+			return fmt.Errorf("wildcard path names are not permitted")
+		}
+
+		// Normalize path string by pruning wildcard keys.
+		pathStr, err := elemToString(elem.Name, elem.Key)
+		if err != nil {
+			return fmt.Errorf("invalid path element: %v", err)
+		}
+
+		// The rank of the child is the number of non-wildcard keys (if any).
```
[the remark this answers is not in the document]
DanG100: there's already a comment on childrenByRank that explains how its used. improved clarity

gnsi/acltrie/trie.go
DanG100: i found it is easier to reason about the func this way. if you have a strong opinion i can change it

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if err := policy.insert(principal, r.GetMode(), r.GetAction()); err != nil {
+		return fmt.Errorf("error inserting policy at %v: %v", r.Path, err)
+	}
+	node.hasPolicy = true
+
+	return nil
+}
+
+// checkPathConflict returns false if the candidate path overlaps with the accepted path.
```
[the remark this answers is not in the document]
DanG100: update comment, this func only looks at a single path element name and keys

gnsi/acltrie/trie.go
wenovus: The suggestion has `[:newLen]` which extends the length as well. Up to you, I think it improves readability.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,199 @@
...
+
+// Trie is the root of the ACL trie.
+type Trie struct {
+	root *trieNode
+}
+
+type trieNode struct {
+	// childrenByRank orders the children of the node by number the non-wildcard keys in the child.
+	// This allows fast lookup of the best match: the one with the largest number of non-wildcard keys.
+	childrenByRank []map[string]*trieNode
```
wenovus, line 36: Suggest to repeat this comment here since people might read this first and ask what a rank is.
`// The rank of the child is the number of non-wildcard keys (if any).`

gnsi/acltrie/trie_test.go
wenovus: Sounds good.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,199 @@
...
+			node = children[pathStr]
+			continue
+		}
+
+		// Before adding a new node, check if the path conflicts with another policy.
+		for _, child := range children {
+			if pathElemsMatch(child.elem, elem) {
+				return fmt.Errorf("policy path conflict with %v/%v", path.Elem[0:i], child.elem)
+			}
+		}
```
wenovus, line 111: Ok, so what this is saying is that `a[foo][*][*]` doesn't conflict with `a[*][bar][baz]` because the latter has a lower rank and should take precedence? Is that gNSI's intended behaviour?

gnsi/acltrie/trie.go
DanG100: done

gnsi/acltrie/trie.go
DanG100: that's right. gNSI spec says to prefer defined key over wildcard element. this implementation works for any number of defined keys.

gnsi/acltrie/trie.go
DanG100: missed that in reading your comment, fixed

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if err := policy.insert(principal, r.GetMode(), r.GetAction()); err != nil {
+		return fmt.Errorf("error inserting policy at %v: %v", r.Path, err)
+	}
+	node.hasPolicy = true
+
+	return nil
+}
+
+// walk explores the trie in breadth first order invoke in the walkFn on every node.
```
wenovus, line 155: typo fix: in breadth first order and invokes walkFn on every node.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,255 @@
...
+// To continue exploring children of the node, the walk func must return true.
```
wenovus, line 156: walk func -> walkFn

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,255 @@
...
+func (t *Trie) walk(walkFn func(node *trieNode, depth int) (bool, error)) error {
+	type traversal struct {
```
wenovus, line 158: suggest traversalNode

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,255 @@
...
+		node  *trieNode
+		depth int
+	}
+
+	tr := []*traversal{{node: t.root}}
```
wenovus, line 163: suggestion: tr -> frontier, or queue

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,255 @@
...
+	for len(tr) > 0 {
+		front := tr[0]
+		tr = tr[1:]
```
wenovus, line 166: I kinda don't like this because I think the underlying array might need to be reallocated in order to free the space at the front. I would prefer a new slice being created as the next frontier. However I could be wrong on this.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,255 @@
...
+	return nil
+}
+
+type compareResult int
+
+const (
+	other compareResult = iota
+	subset
+	superset
+)
```
wenovus, line 186: I think it would make the code more readable if we split other into `equal` and `disjoint`, and then can just switch on the result. (which would incidentally make it suitable to upstream to ygot)

wenovus (changes requested):

gnsi/acltrie/trie.go
DanG100: done

gnsi/acltrie/trie.go
DanG100: done

gnsi/acltrie/trie.go
DanG100: done

gnsi/acltrie/trie.go
DanG100: queue

gnsi/acltrie/trie.go
DanG100: this seems to be the standard way of implementing queues. Removing the the first elem reduces the capacity of the list 1.

gnsi/acltrie/trie.go
DanG100: agree a more generic compare utility would be nice. i've added a todo for this, because needs some extra thought to handle paths of different lengths, wildcards in the name, etc.

gnsi/acltrie/trie.go
wenovus: Sounds good to me, it *looks* like Go will garbage collect the beginning of the slice that's unreferenceable.

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,256 @@
...
+// error: not two keys are both subset and superset.
+// TODO: Make to more broadly useful compare that outputs subset, superset, equal, disjoint.
+func comparePathElem(a, b *gpb.PathElem) (compareResult, error) {
+	result := other
+	for k, aVal := range a.Key {
+		bVal, ok := b.Key[k]
+		switch {
+		case aVal == bVal:
+			continue
+		case aVal == "*" && !ok:
```
wenovus, line 201: annotate `// b is implicitly wildcarded`

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+// comparePathElem compare two path elements a, b and returns:
+// subset: if every definite key in a is wildcard in b.
+// superset: if every wildcard key in b is non-wildcard in b.
+// other: all keys are the same or all keys are different.
+// error: not two keys are both subset and superset.
+// TODO: Make to more broadly useful compare that outputs subset, superset, equal, disjoint.
+func comparePathElem(a, b *gpb.PathElem) (compareResult, error) {
+	result := other
```
wenovus, line 195: result -> setRelation

gnsi/acltrie/trie.go
```diff
@@ -0,0 +1,256 @@
...
+			result = subset
+		}
+	}
+	for k, bVal := range b.Key {
+		_, ok := a.Key[k]
+		if !ok && bVal != "*" { // If a contains an implicit wildcard.
+			if result == subset {
+				return other, fmt.Errorf("path %v is not consistently a superset of %v", a, b)
+			}
+		}
```
wenovus, line 221: missing `result = superset` at end of outer if statement? A missing test case?

wenovus (approved):

gnsi/acltrie/trie.go
DanG100: done and added case

PR #142, opened 2023-02-15
Implement Probe for the acltrie

DanG100:


wenovus (commented):
Just a first look.

gnsi/acltrie/trie.go
```diff
@@ -152,6 +175,70 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
 	return nil
 }
 
+// Probe returns the for the given path, user, and mode; if there is no match, deny is returned
```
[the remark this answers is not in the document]
DanG100: Yes. It's for the probe RPC, it's returns of user is allowed to write/read to the path

gnsi/acltrie/trie.go
```diff
@@ -255,3 +342,37 @@ func elemToString(name string, kv map[string]string) (string, error) {
 
 	return name, nil
 }
+
+// pathElemsMatch returns true if the a PathElem matches the b PathElem.
+// A match is when both the name match and all keys match (where * or unset matches with anything).
+func pathElemsMatch(a, b *gpb.PathElem) bool {
```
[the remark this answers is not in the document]
DanG100: i have in pr in ygot for this

gnsi/acltrie/trie.go
```diff
@@ -27,7 +27,8 @@ import (
 
 // Trie is the root of the ACL trie.
 type Trie struct {
-	root *trieNode
+	root        *trieNode
+	memberships map[string]map[string]bool
```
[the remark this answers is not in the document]
DanG100: group name to user set, added comment

DanG100:
I've simplified the logic with the ygot compare func. PTAL @wenovus

gnsi/acltrie/trie.go
```diff
@@ -39,7 +42,29 @@ type trieNode struct {
...
+		}
+	}
+
+	var act pathzpb.Action
+	for group, action := range tn.groups[mode] {
+		if _, ok := memberships[group][user]; ok {
+			if action == pathzpb.Action_ACTION_DENY { // DENY action take precedence over PERMIT.
+				return action, false
+			}
+			act = action
```
wenovus, line 61: technically this should be a switch case with a default statement returning error in case we add another enum value.

gnsi/acltrie/trie.go
```diff
@@ -152,79 +165,104 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
...
+		last := stack[len(stack)-1]
+		stack = stack[:len(stack)-1]
+		for _, c := range last.node.children {
+			// Handle inconsistency with the start of a path being at depth 1,
+			// but should be 0th elem in the slice.
+			if last.depth == 1 {
+				path.Elem = []*gpb.PathElem{last.node.elem}
+			}
+			path.Elem = append(path.Elem[:last.depth], c.elem)
+			cont, err := walkFn(c, path)
```
wenovus, line 254: I believe the reason for the inconsistency is due to the `walkFn` being invoked for the children *before* they're added to the stack as opposed to *directly after* popping from the stack. This is causing the nodes to be visited in a non-DFS order, which is what is causing the issue. If `walkFn` is called prior to the inner loop, then the visiting will occur in DFS order, and then the path would be consistent.

An implication is that now the root node will be visited, and the current algorithm doesn't visit the root.

wenovus (changes requested):

gnsi/acltrie/trie.go
```diff
@@ -106,29 +131,17 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
...
-				if pathCmp != other && elemCmp != pathCmp {
-					return false, fmt.Errorf("path is not consistently subset or superset of other rules")
-				}
-				if elemCmp != other {
-					pathCmp = elemCmp
-				}
-
-				n = n.parent
+		if len(walkPath.Elem) == len(path.Elem) && node.hasPolicy {
+			fmt.Println(walkPath)
```
wenovus, line 142: debug statement?

gnsi/acltrie/trie.go
```diff
@@ -152,79 +165,104 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
 	return nil
 }
 
-// walk explores the trie in breadth first order and invokes walkFn on every node.
+// Probe returns the action for the given path, user, and mode; if there is no match, deny is returned
+func (t *Trie) Probe(path *gpb.Path, user string, mode pathzpb.Mode) pathzpb.Action {
+	potentialPolicies := []*trieNode{}
+	maxLen := 0
```
wenovus, line 171: longestPolicyLen

gnsi/acltrie/trie.go
```diff
@@ -152,79 +165,104 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
 	return nil
 }
 
-// walk explores the trie in breadth first order and invokes walkFn on every node.
+// Probe returns the action for the given path, user, and mode; if there is no match, deny is returned
+func (t *Trie) Probe(path *gpb.Path, user string, mode pathzpb.Mode) pathzpb.Action {
+	potentialPolicies := []*trieNode{}
```
wenovus, line 170: matchingPolicies?

gnsi/acltrie/trie.go
```diff
@@ -152,79 +165,104 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
...
+		for i := maxLen - 1; i >= 0; i-- {
+			path.Elem[i] = n.elem
+			setKey += setKeys(path.Elem[i].Key)
+			n = p.parent
+		}
+		if setKey >= maxSetKeys {
+			maxSetKeys = setKey
+			bestPaths = append(bestPaths, p)
+		}
+	}
```
wenovus, line 213: I wonder if this can be cached in the trie during insertion so there isn't a need for this block?

gnsi/acltrie/trie.go
```diff
@@ -152,79 +165,104 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
...
+		}
+
+		return true, nil
+	})
+	if len(potentialPolicies) == 0 {
+		return pathzpb.Action_ACTION_DENY
+	}
+
+	// Pick the policies with the largest number of definite keys.
+	var bestPaths []*trieNode
```
wenovus, line 196: mostSpecificPolicies

gnsi/acltrie/trie.go
DanG100: done

gnsi/acltrie/trie.go
DanG100: added check on insert

wenovus (approved):

PR #144, opened 2023-02-22
Implement the pathz RPC server

DanG100:
Note: i removed the path checking because the rule prefer DENY over PERMIT handles those cases


gnsi/acltrie/trie.go
```diff
@@ -134,25 +156,6 @@ func (t *Trie) Insert(r *pathzpb.AuthorizationRule) error {
 
 		node = node.children[pathStr]
 	}
-	// Validate the path by ensuring that compared to all other paths with the same length and same name,
-	// all list keys do not overlap.
-	err := t.walk(func(node *trieNode, walkPath *gpb.Path) (bool, error) {
```
wenovus, line 139: Should we not error out here to avoid a bad policy from being installed?

DanG100: Since Probe prefers DENY over ALLOW, these policies are ok to install, since there's no ambiguity.

wenovus: Actually I don't see where in the README where DENY is preferred over ALLOW. If that is the case, then we're essentially saying this whole section doesn't exist, since there will never be a conflict? Since if there is overlapping policies then we will always say DENY if there is a match?

https://github.com/openconfig/gnsi/blob/main/pathz/authorization-README.md#conflict-resolution

Conflict Resolution
Policy evaluation should end with a single best match rule for the provided user / path / mode. If there is more than one 'best match', an error must be logged and evaluation should return a failure.
...
Authorization rules must be defined such that a single best match is possible. If the result of policy evaluation is more than one match, an error must be raised.

gnsi/pathz/pathz.go
```diff
@@ -0,0 +1,158 @@
...
+		case *pathzpb.RotateRequest_FinalizeRotation:
+			if !receivedUploadReq {
+				return status.Error(codes.FailedPrecondition, "finalize rotation called before upload request")
+			}
+			s.activeMu.Lock()
+			s.sandboxMu.Lock()
+			s.active = s.sandbox
+			s.sandbox = nil
+			s.activeMu.Unlock()
+			s.sandboxMu.Unlock()
```
wenovus, line 93: lock and unlock in nested order to avoid deadlock.

gnsi/pathz/pathz.go
```diff
@@ -0,0 +1,158 @@
...
+			if err := rs.Send(&pathzpb.RotateResponse{}); err != nil {
+				return err
+			}
+			return nil
+		}
+	}
+}
+
+func (s *Server) getPolicyWithLock(i pathzpb.PolicyInstance) (*policyData, *sync.RWMutex, error) {
```
wenovus, line 102: withRLock

gnsi/acltrie/trie_test.go
```diff
@@ -266,96 +266,6 @@ func TestInsert(t *testing.T) {
...
-	}, {
-		desc: "failure ambiguous list keys",
-		initialRule: &pathzpb.AuthorizationRule{
-			Path:      mustPath("/foo[a=*][b=2]"),
-			Principal: &pathzpb.AuthorizationRule_User{User: "bob"},
-			Mode:      pathzpb.Mode_MODE_READ,
-			Action:    pathzpb.Action_ACTION_PERMIT,
-		},
-		testRule: &pathzpb.AuthorizationRule{
-			Path:      mustPath("/foo[a=1][b=*]"),
```
wenovus, line 278: Add a test with wildcards of same path conflicting.

wenovus (approved):

gnsi/acltrie/trie.go
DanG100: working on readme update for this

PR #151, opened 2023-03-03
Support adding initial connected routes configuration to lemming.

wenovus:
This is useful when doing standalone testing where having initial connected routes is necessary (e.g. gRIBI testing)

DeviceOpt was added. This PR is mostly copied from https://github.com/openconfig/gribigo/blob/main/device/device.go

### DIFF
--- a/lemming.go
+++ b/lemming.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openconfig/lemming/dataplane"
 	fgnmi "github.com/openconfig/lemming/gnmi"
 	"github.com/openconfig/lemming/gnmi/fakedevice"
+	"github.com/openconfig/lemming/gnmi/oc"
 	"github.com/openconfig/lemming/gnmi/reconciler"
 	fgnoi "github.com/openconfig/lemming/gnoi"
 	fgnsi "github.com/openconfig/lemming/gnsi"
@@ -33,6 +34,7 @@ import (
 	"github.com/openconfig/lemming/sysrib"
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/reflection"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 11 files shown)


lemming.go
```diff
@@ -55,8 +57,56 @@ type Device struct {
...
+func TLSCreds(c credentials.TransportCredentials) *tlsCreds {
+	return &tlsCreds{c: c}
+}
+
+// IsDevOpt implements the DevOpt interface for tlsCreds.
+func (*tlsCreds) isDevOpt() {}
+
 // New returns a new initialized device.
-func New(lis net.Listener, targetName, zapiURL string, opts ...grpc.ServerOption) (*Device, error) {
+func New(lis net.Listener, targetName, zapiURL string, opts ...DevOpt) (*Device, error) {
```
DanG100, line 109: i prefer how ygnmi does options, the code is much simpler. 

https://github.com/openconfig/ygnmi/blob/main/ygnmi/ygnmi.go#L171

wenovus: Done.

lemming.go
```diff
@@ -55,8 +57,56 @@ type Device struct {
...
+func TLSCredsFromFile(certFile, keyFile string) (*tlsCreds, error) {
+	t, err := credentials.NewServerTLSFromFile(certFile, keyFile)
+	if err != nil {
+		return nil, err
+	}
+	return &tlsCreds{c: t}, nil
+}
+
+// TLSCreds returns a wrapper of TransportCredentials into a DevOpt.
+func TLSCreds(c credentials.TransportCredentials) *tlsCreds {
```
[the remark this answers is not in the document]
wenovus: Done.

lemming.go
```diff
@@ -55,8 +57,57 @@ type Device struct {
...
+	}
+	return o
+}
+
+// WithDeviceConfig sets the startup config of the device to c.
+// Today we do not allow the configuration to be changed in flight, but this
+// can be implemented in the future.
+//
+// This DeviceOption is intended for standalone device testing.
+func WithDeviceConfig(c []byte) Option {
```
DanG100, line 84: nit: WithInitialConfig?

lemming.go
```diff
@@ -70,7 +121,26 @@ func New(lis net.Listener, targetName, zapiURL string, opts ...grpc.ServerOption
 		recs = append(recs, dplane)
 	}
 
-	s := grpc.NewServer(opts...)
+	resolvedOpts := resolveOpts(opts)
+
+	jcfg := resolvedOpts.deviceConfigJSON
+	root := &oc.Root{}
+	switch jcfg {
```
DanG100, line 128:
```
if jcfg := resolvedOpts.deviceConfigJSON; jcfg != nil {
 unmarshal
} 
```

lemming.go
```diff
@@ -70,7 +121,26 @@ func New(lis net.Listener, targetName, zapiURL string, opts ...grpc.ServerOption
...
+	case nil:
+		root.GetOrCreateNetworkInstance(fakedevice.DefaultNetworkInstance).Type = oc.NetworkInstanceTypes_NETWORK_INSTANCE_TYPE_DEFAULT_INSTANCE
```
DanG100, line 130: do we always want to create a default network instance, if even when a config is specified? i'm not sure how the initial is intended to be used

DanG100 (changes requested):

lemming.go
```diff
@@ -55,8 +57,57 @@ type Device struct {
...
+	if err != nil {
+		return nil, err
+	}
+	return func(o *opt) {
+		o.tlsCredentials = t
+	}, nil
+}
+
+// WithTLSCreds returns a wrapper of TransportCredentials into a DevOpt.
+func WithTLSCreds(c credentials.TransportCredentials) Option {
```
DanG100, line 103: nit: if insecure.Credentials are meant to be used with the option then, call it WithCreds() or WithTransportCreds

lemming.go
wenovus: The initial config may specify a different-named network instance, which is why this is guarded.

lemming.go
wenovus: Done.

lemming.go
wenovus: Done.

lemming.go
wenovus: Done.

lemming.go
wenovus: There should be one and only one default network instance.

DanG100 (approved):

PR #152, opened 2023-03-04
Change lemming's g* services to run on different ports per intent.

wenovus:
- gNMI/gNOI/gNSI on one port
- gRIBI on another
- P4RT on another

These were intended to run on different ports.

NOTE: There is actually a hack in here where I made the gRIBI service actually register in the same place as gNMI/gNOI. This is to deal with the fact that KNE doesn't have lemming registered with the right port. This can be changed after that is done.

### DIFF
--- a/lemming.go
+++ b/lemming.go
@@ -74,16 +74,9 @@ type opt struct {
 	deviceConfigJSON []byte
 	// tlsCredentials contains TLS credentials that can be used for a device.
 	tlsCredentials credentials.TransportCredentials
-	gribiAddr      serviceAddr
-	gnmiAddr       serviceAddr
-	p4rtAddr       serviceAddr
-}
-
-// serviceAddr is the internal implementation that specifies the port that a g*
-// service should listen on.
-type serviceAddr struct {
-	host string
-	port int
+	gribiAddr      string
+	gnmiAddr       string
+	p4rtAddr       string
 }
 
 // resolveOpts applies all the options and returns a struct containing the result.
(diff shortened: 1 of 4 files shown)


lemming.go
```diff
@@ -87,6 +106,60 @@ func WithDeviceConfig(c []byte) Option {
...
+
+// WithP4RTAddr is a device option that specifies that the P4RT address.
+func WithP4RTAddr(host string, port int) Option {
+	return func(o *opt) {
+		o.p4rtAddr.host = host
+		o.p4rtAddr.port = port
+	}
+}
+
+// gNMIAddress is the internal implementation that specifies the port that gNMI should
```
DanG100, line 133: all this can be removed?

lemming.go
```diff
@@ -87,6 +106,60 @@ func WithDeviceConfig(c []byte) Option {
 	}
 }
 
+// WithGRIBIAddr is a device option that specifies that the gRIBI address.
+func WithGRIBIAddr(host string, port int) Option {
```
DanG100, line 110: it seems like we could just be using a string everywhere, there no reason to separate host and port, unless i'm missing something

lemming.go
wenovus: Done

lemming.go
wenovus: Done

wenovus:
Note I created a PR in KNE to change the ports -- recompiling kne_cli alone wasn't enough to fix this issue, need your help here @DanG100 : https://github.com/openconfig/kne/pull/318

DanG100 (approved):

PR #153, opened 2023-03-06
Update AFT paths as a gRIBI hook function

wenovus:
This is a temporary solution since it doesn't handle deletes.

This would support integrating lemming into gribigo

### DIFF
--- a/gribi/gribi.go
+++ b/gribi/gribi.go
@@ -24,11 +24,17 @@ import (
 	"github.com/openconfig/gribigo/afthelper"
 	"github.com/openconfig/gribigo/constants"
 	"github.com/openconfig/gribigo/server"
+	"github.com/openconfig/lemming/gnmi/gnmiclient"
+	"github.com/openconfig/lemming/gnmi/oc"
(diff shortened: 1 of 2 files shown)


gribi/gribi.go
```diff
@@ -79,6 +85,28 @@ func createGRIBIServer(root *oc.Root) (*server.Server, error) {
...
+			// TODO(wenbli): handle replace and delete :-)
+			return
+		}
+		// write gNMI notifications
+		if err := updateAft(yclient, o, ni, data); err != nil {
+			log.Errorf("invalid notifications, %v", err)
+		}
+
+		// server.WithFIBProgrammedCheck()
+		//   -> gives us a function that checks whether an ID is a tristate (ok, failed, pending)
```
DanG100, line 104: i'm not sure what this comment block is for. either add a todo or remove

DanG100 (approved):

gribi/gribi.go
wenovus: Added

PR #155, opened 2023-03-08
Skip validating state SetRequest/updates.

wenovus:
Note that replace/deletes are not affected.

This is to enable higher performance (e.g. gRIBI adding 1000 routes).

This is a temporary solution to enable gribigo to start using lemming. The long term solution is to create an UnmarshalSetRequest utility in ygot to output all the deleted/updated paths. This would enable an O(size of SetRequest) way of computing the diff, as well as reverting the SetRequest should the validation fail.

### DIFF
--- a/gnmi/gnmi.go
+++ b/gnmi/gnmi.go
@@ -242,11 +242,32 @@ func updateCacheNotifs(cache *cache.Cache, nos []*gpb.Notification, target, orig
 			}
 			pathsForDelete = append(pathsForDelete, p)
 		}
-		log.V(1).Infof("datastore: updating the following values: %+v", n.Update)
-		log.V(1).Infof("datastore: deleting the following paths: %+v", pathsForDelete)
+		if len(n.Update) > 0 {
+			log.V(1).Infof("datastore: updating the following values: %+v", n.Update)
+		}
+		if len(pathsForDelete) > 0 {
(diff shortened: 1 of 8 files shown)


gnmi/gnmi.go
```diff
@@ -260,6 +281,30 @@ func updateCacheNotifs(cache *cache.Cache, nos []*gpb.Notification, target, orig
 // the gNMI cache.
 // - auth adds authorization to before writing vals to the cache, if set to nil, not authorization is checked.
 func set(schema *ytypes.Schema, cache *cache.Cache, target string, req *gpb.SetRequest, preferShadowPath bool, validators []func(*oc.Root) error, timestamp int64, user string, auth PathAuth) error {
+	// skip validation and diffing for performance.
+	// TODO(wenbli): This if block should be removed once a more performant diffing utility is created.
+	if !preferShadowPath && len(req.Delete)+len(req.Replace) == 0 {
```
DanG100, line 286: i thought we didn't need to validate state paths at all?

DanG100 (approved):

gnmi/gnmi.go
wenovus: replace/delete don't work because they affect the config paths as well. The longterm solution is needed for those cases.

PR #157, opened 2023-03-10
Remove recursive locking in sysrib and add `go test -race`

wenovus:
Unfortunately recursive locking (not supported in Go) is not something the Go team believes is interesting to detect because the interesting cases are not actually detectable:
https://github.com/golang/go/issues/35138, so there is no CI tool for it.

The dataplane and gnmi packages currently fail the data race tests.


wenovus:
The issue was `ResolveAndProgramDiff` was calling RLock(), and then it was recursively calling entryForCIDR which also called RLock(). If some other thread calls Lock() prior to the second RLock(), which in this case was AddRoute, then the second RLock() will block to avoid starvation of the write lock, and that caused a deadlock.

DanG100 (approved):

PR #162, opened 2023-03-17
Support BGP-triggered GUE for IPv4-mapped IPv6 routes

wenovus:
* Put all of v4, v6, and v4-mapped v6 test cases into one file.
* Reduced traffic generation time to lower running time of GUE test.
* Fixed sysrib's ZAPI route exchange so that it's the resolved route's nexthops that are exchanged to ZAPI clients.


dataplane/forwarding/infra/fwdpacket/packet.go
```diff
@@ -42,6 +42,10 @@ const (
 	LogDesc                // sets the packet's description (prefixed to all messages)
 )
 
+const (
```
DanG100, line 45: this isn't right way to turn off the the trace, it's being turned on here https://github.com/openconfig/lemming/blob/main/dataplane/forwarding/fwdport/ports/kernel.go#L110

wenovus: Oops removed. Added common debug package to put the debug flags.

dataplane/forwarding/fwdport/ports/kernel.go
```diff
@@ -107,7 +108,7 @@ func (p *kernelPort) process() {
 					fwdport.Increment(p, len(pkt.Data()), fwdpb.CounterId_COUNTER_ID_RX_BAD_PACKETS, fwdpb.CounterId_COUNTER_ID_RX_BAD_OCTETS)
 					continue
 				}
-				fwdPkt.Debug(true) // TODO: put this behind a flag
+				fwdPkt.Debug(debug.PacketTrace)
```
DanG100, line 111: i believe debug is turned in the tap interface code as well

DanG100 (approved):

PR #164, opened 2023-03-21
Change BGP port on ATE to port2

wenovus:
The nexthop of the BGP route is port2, so the route should come from there as well instead of port1.


DanG100:
Is this passing locally for you? Looks the bgp triggered gue is failing on presubmit? Can we get it passing or remove from presubmit

wenovus:
@DanG100 This is passing for me locally. It looks like in the VM, or higher CPU environment there is more packet loss. For now I've tweaked the loss tolerance higher and marked a TODO for later debugging since otherwise the test is working fine.

DanG100 (approved):

PR #171, opened 2023-03-31
Refactor ygnmi test

wenovus:
Reuse checking logic

### DIFF
--- a/gnmi/gnmi_test.go
+++ b/gnmi/gnmi_test.go
@@ -613,7 +613,7 @@ func TestSetYGNMI(t *testing.T) {
 		desc      string
 		isState   bool
 		inOp      func(c *ygnmi.Client) error
-		getOp     func(t *testing.T, c *ygnmi.Client) interface{}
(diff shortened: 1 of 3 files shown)


gnmi/gnmi_test.go
```diff
@@ -909,8 +867,32 @@ func TestSetYGNMI(t *testing.T) {
...
+				return
+			}
+
+			got, ok := tt.getOp(t, c)
+			switch want := tt.wantValue.(type) {
+			case nil:
+				if ok {
+					t.Errorf("Got present, want not present")
+				}
+			case ygot.GoStruct:
```
DanG100, line 880: is there a benefit to doing ygot.Diff vs cmp.DIff here?

DanG100 (approved):

gnmi/gnmi_test.go
wenovus: cmp.Diff doesn't work because it treats nil vs. empty structures as different.

PR #176, opened 2023-04-27
Create placeholders for sai implementation

DanG100:
* Renamed dataplane service because it conflicts with c++ generated code
* Create standalone dataplane
* From go start gRPC server
* From c++ translate sai calls and make gRPC requests


dataplane/standalone/sai_entrypoint.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "translator.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// TODO: implement this without using gRPC.
+sai_status_t sai_api_initialize(_In_ uint64_t flags, _In_ const sai_service_method_table_t *services)
+{
+    initialize(GoInt(50000));
```
wenovus, line 24: Also include lucius_clib.h, sai.h, and the forwarding proto library -- don't rely on transitive includes to allow translator.h to change its own includes.

dataplane/standalone/translator.cc
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "translator.h"
+
+using namespace Translator;
+
+sai_status_t Translator::create_switch(_Out_ sai_object_id_t *switch_id, _In_ uint32_t attr_count, _In_ const sai_attribute_t *attr_list)
+{
+    return SAI_STATUS_SUCCESS;
```
wenovus, line 21: nit: Likewise here include "sai.h" since you're using definitions from it.

dataplane/standalone/translator.h
```diff
@@ -0,0 +1,37 @@
...
+#include "dataplane/standalone/lucius/lucius_clib.h"
+
+extern "C"
+{
+#include "sai.h"
+}
+
+namespace Translator {
+    sai_status_t create_switch(_Out_ sai_object_id_t *switch_id, _In_ uint32_t attr_count, _In_ const sai_attribute_t *attr_list);
+    extern std::unique_ptr<forwarding::Forwarding::Stub> client;
```
wenovus, line 34: Why is this exported?

dataplane/standalone/sai_entrypoint.cc
```diff
@@ -0,0 +1,54 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "translator.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
wenovus, line 19: Do you have documentation on why this is needed?

dataplane/standalone/sai_entrypoint.cc
wenovus, line 1: My understanding is that this is what will compile with sai.h, is that right?

dataplane/standalone/translator.h
DanG100: changed to class

dataplane/standalone/sai_entrypoint.cc
DanG100: the build rules produce a shared libraries (libsai.so) that implements the funcs in the sai.h. The users are excepted to have their own copy sai.h

dataplane/standalone/sai_entrypoint.cc
DanG100: i think i just need these on includes. It's so that the c++ funcs are callable from c

dataplane/standalone/sai_entrypoint.cc
```diff
@@ -0,0 +1,63 @@
...
+                                  grpc::InsecureChannelCredentials());
+  translator = std::make_shared<Translator>(chan);
+  return SAI_STATUS_SUCCESS;
+}
+
+sai_status_t sai_api_query(_In_ sai_api_t api, _Out_ void **api_method_table) {
+  switch (api) {
+    case SAI_API_SWITCH: {
+      sai_switch_api_t *swapi = reinterpret_cast<sai_switch_api_t *>(
+          malloc(sizeof(sai_switch_api_t)));
```
wenovus, line 50: Since we're doing a malloc here, add TODO or in this PR implement sai_api_uninitialize to deallocate the memory.

wenovus (approved):

dataplane/standalone/sai_entrypoint.cc
DanG100: done

PR #179, opened 2023-05-04
Create stubs for all sai APIs

DanG100:
* Everything under dataplane/standalone/sai is autogenerated
* Minor fix to WORKSPACE to load packages in the write order
* Included code generator


dataplane/standalone/translator.h
```diff
@@ -20,14 +20,14 @@
...
 #include "proto/forwarding/forwarding_service.pb.h"
 
 extern "C" {
 #include "inc/sai.h"
 }
 
+
 class Translator {
  public:
   explicit Translator(std::shared_ptr<grpc::Channel> chan) {
```
wenovus, line 33: Is the `create_switch` function here still going to be used to populate the tables by overwriting the stub from the generated code?

dataplane/standalone/apigen/apigen.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+		}
+
+	}
+	return &sa
+}
+
+var (
+	headerTmpl = template.Must(template.New("header").Parse(`
+// Copyright 2023 Google LLC
```
wenovus, line 142: extra newline

dataplane/standalone/apigen/apigen.go
```diff
@@ -0,0 +1,303 @@
...
+extern const {{ .APIType }} l_{{ .APIName }};
+
+{{ range .Funcs }}
+{{ .ReturnType }} l_{{ .Name }}({{ .Args }});
+{{ end }}
+
+#endif  // {{ .IncludeGuard }}
+`))
+	ccTmpl = template.Must(template.New("cc").Parse(`
+// Copyright 2023 Google LLC
```
wenovus, line 172: extra newline

wenovus (approved):

dataplane/standalone/BUILD.bazel
```diff
@@ -4,21 +4,33 @@ cc_library(
...
     linkshared = True,
+    linkstatic = True,
+    deps = [
+        ":saigo",
+    ],
+)
+
+cc_binary(
+    name = "saibin",
+    linkstatic = True,
```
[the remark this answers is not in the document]
DanG100: it helps find some linker issues because it's not a shared library. In the shared library, missing libraries won't cause compile time errors.

dataplane/standalone/translator.h
DanG100: the stub will call this func is my plan, to implement the sai API we need global functions, but those would be harder to test than class methods. So the stubs are going to call the methods on a global instance of the translator class

PR #181, opened 2023-05-08
Implement switch create and basic object and attribute tracking

DanG100:


dataplane/standalone/translator.h
```diff
@@ -19,28 +19,45 @@
...
+class Switch;
+class Port;
 
 class Translator {
  public:
   explicit Translator(std::shared_ptr<grpc::Channel> chan) {
     client = forwarding::Forwarding::NewStub(chan);
+    objects.push_back(SAI_OBJECT_TYPE_NULL);  // ID == 0, is invalid so skip.
+    sw = std::make_unique<Switch>(this, std::move(client));
+    port = std::make_unique<Port>(this, std::move(client));
```
wenovus, line 44: I'm not sure about the usage of std::move. Usually std::move can leave the argument in a "valid but unspecified state": https://en.cppreference.com/w/cpp/utility/move

Here could client be a shared_ptr so that you can pass it in to different constructors?

dataplane/standalone/sai/common.h
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+#ifndef DATAPLANE_STANDALONE_SAI_COMMON_H_
+#define DATAPLANE_STANDALONE_SAI_COMMON_H_
+
+#include <memory>
+
+#include "dataplane/standalone/translator.h"
+
+extern std::shared_ptr<Translator> translator;
```
wenovus, line 22: Why was this moved from translator.h to here?

dataplane/standalone/switch.cc
```diff
@@ -0,0 +1,226 @@
...
+sai_status_t Switch::create_switch(_Out_ sai_object_id_t *switch_id,
+                                   _In_ uint32_t attr_count,
+                                   _In_ const sai_attribute_t *attr_list) {
+  *switch_id = this->translator->createObject(SAI_OBJECT_TYPE_SWITCH);
+  for (uint32_t i = 0; i < attr_count; i++) {
+    LOG(attr_list[i].id);
+    this->translator->setAttribute(*switch_id, sai_attribute_t{
+                                                   .id = attr_list[i].id,
+                                                   .value = attr_list[i].value,
+                                               });
```
wenovus, line 29: There is inconsistent formatting of these function calls. I think I prefer the one argument per line style (e.g. //36).

dataplane/standalone/switch.cc
```diff
@@ -0,0 +1,226 @@
...
+    switch (attr_list[i].id) {
+      // TODO(dgrau): handle this attributes specially.
```
wenovus, line 31: Is this switch statement doing something other than just casting? Is this what this TODO is about?

dataplane/standalone/port.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "dataplane/standalone/translator.h"
+#include "proto/forwarding/forwarding_service.grpc.pb.h"
+#include "proto/forwarding/forwarding_service.pb.h"
+
+extern "C" {
+#include "inc/sai.h"
+}
+
+class Translator;
```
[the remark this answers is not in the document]
DanG100: the forward declaration is needed because the class depend on each other.

dataplane/standalone/switch.cc
DanG100: this what the formatter does to get <80 chars per line (per google style guide)

dataplane/standalone/switch.cc
DanG100: that's right, its not currently doing anything. these attribute need to dealt with in a future PR

dataplane/standalone/log/log.h
```diff
@@ -24,4 +24,6 @@ extern std::ofstream logger;
          << std::endl;                                               \
   logger.flush()
 
+#define LOG(msg) logger << "Line: " << msg << std::endl; logger.flush()
```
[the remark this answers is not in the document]
DanG100: done

dataplane/standalone/translator.h
```diff
@@ -19,28 +19,45 @@
...
+  void setAttribute(sai_object_id_t id, sai_attribute_t attr);
+  sai_status_t getAttribute(sai_object_id_t id, sai_attribute_t *attr);
+
+  std::unique_ptr<Switch> sw;
+  std::unique_ptr<Port> port;
 
  private:
   std::unique_ptr<forwarding::Forwarding::Stub> client;
+  // objects maintains a global list of object id to types.
+  std::vector<sai_object_type_t> objects;
```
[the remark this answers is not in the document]
wenovus: I'm wondering if it makes sense to merge the vector and the map since their keys are the same? Are you using the vector as a map?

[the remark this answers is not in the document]
DanG100: done

wenovus (approved):

PR #182, opened 2023-05-09
BGP route propagation unit test with preliminary Loc-RIB population.

wenovus:
test helper code copied from existing integration tests. I plan to clean this up after the form of these unit tests are more established.


DanG100 (approved):

DanG100:
Can you looking to the failing test?

wenovus:
> Can you looking to the failing test?

The race is happening within GoBGP itself. The zebra client within the BGP server will read the `neighborMap` when propagating newly received routes, but `StopBGP` will delete the neighbours on shutdown. I don't think this is an issue that warrants a fix right now since it only happens during shutdown but I can mark a TODO so that we're aware why we turned off race tests for local_tests.

PR #186, opened 2023-05-12
Fix propagation of BGP routes through a device in `local_tests`.

wenovus:
It didn't work before because GoBGP did not recognize loopback nexthop IP addresses as valid. So I updated it to allow loopback addresses when the RouterID is set to a loopback address so we can propagate routes correctly. GoBGP doesn't do other checks with loopback addresses.

https://github.com/wenovus/gobgp/pull/1


WORKSPACE
```diff
@@ -230,6 +230,13 @@ go_repository(
     version = "v2.2.6",
 )
 
+go_repository(
+    name = "com_github_osrg_gobgp_v3",
+    importpath = "github.com/osrg/gobgp/v3",
+    sum = "h1:zfXhM8+QSfAPUkzWYTbH8pMkvxkcUfvQEKnBU7ikk4g=",
```
DanG100, line 236: don't add go_repository to this file, add them to reposirories.bzl

DanG100 (approved):

WORKSPACE
wenovus: Updated Makefile to write to this file.

PR #187, opened 2023-05-15
Finish refactor for api

DanG100:
* Create common attribute tracking class
* Switch class contains other class with other apis
* Create and set operator are delegated to classes
* SAI.patch contains autogenerated code that i couldn't generate with bazel. Will investigate in future PR


dataplane/standalone/translator.cc
```diff
@@ -83,25 +112,56 @@ sai_status_t Translator::remove(sai_object_type_t type, common_entry_t id) {
...
+                                       sai_attribute_t* attr_list) {
+  return this->attrMgr->get_attribute(std::to_string(id), attr_count,
+                                      attr_list);
 }
 
 sai_status_t Translator::get_attribute(sai_object_type_t type,
                                        common_entry_t id, uint32_t attr_count,
                                        sai_attribute_t* attr_list) {
-  return sai_status_t();
+  std::string idStr = this->attrMgr->serialize_entry(type, id);
```
wenovus, line 163: (no action required) What does serialize_entry do here? It extracts the sai_object_id out of the sai_entry_t? The name doesn't seem to indicate that.

dataplane/standalone/translator.h
```diff
@@ -132,8 +114,10 @@ class Translator {
                                   sai_status_t *object_statuses);
 
  private:
+  std::shared_ptr<AttributeManager> attrMgr;
   std::shared_ptr<forwarding::Forwarding::Stub> client;
-  std::unordered_map<sai_object_id_t, SaiObject> objects;
+  std::unordered_map<sai_object_id_t, std::shared_ptr<Switch>> switches;
+  std::unordered_map<std::string, std::shared_ptr<APIBase>> apis;
```
wenovus, line 120: Doesn't the switch also contain the object apis (IIUC)? Are there apis that are non-switch apis?

dataplane/standalone/translator.cc
```diff
@@ -83,25 +112,56 @@ sai_status_t Translator::remove(sai_object_type_t type, common_entry_t id) {
...
+  if (switch_id != SAI_NULL_OBJECT_ID) {
+    status = this->switches[switch_id]->set_child_attr(type, idStr, attr);
+  } else {
+    status = this->apis[idStr]->set_attribute(attr);
+  }
+  if (status != SAI_STATUS_SUCCESS) {
+    return status;
+  }
+
+  this->attrMgr->set_attribute(idStr, attr);
```
wenovus, line 148: (no action required) Is it intended to call set_attribute twice?

dataplane/standalone/translator.cc
```diff
@@ -14,62 +14,91 @@
...
+      attr_list);  // Delegate creation to switch instance.
+  if (status != SAI_STATUS_SUCCESS) {
+    return status;
+  }
+  // Save attributes to attr managers.
+  for (uint32_t i = 0; i < attr_count; i++) {
+    this->attrMgr->set_attribute(idStr, attr_list[i]);
+  }
+
+  return SAI_STATUS_SUCCESS;
```
wenovus, line 101: This looks really similar to the function above. Can call into the other?

dataplane/standalone/translator.cc
DanG100: some object are referred to by integer id (ports, switches, etc) while other use are tuples (routes are tuple of switch id, virtual router id, ip prefix). To have a common data model, i use a string as key, the string is either the object id or the stringified tuple.

dataplane/standalone/common.cc
```diff
@@ -0,0 +1,153 @@
...
+      .type = type,
+      .switch_id = switch_id,
+      .attributes = std::unordered_map<sai_attr_id_t, sai_attribute_value_t>(),
+  };
+  return id;
+}
+
+void AttributeManager::create(sai_object_type_t type, std::string id,
+                              sai_object_id_t switch_id) {
+  this->objects[id] = {
```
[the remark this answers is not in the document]
DanG100: no, object ids are unique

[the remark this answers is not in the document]
wenovus: Oh so the `id` fed into `create` is already created by another id allocator that ensures correctness?

dataplane/standalone/translator.cc
DanG100: yes, one call is one the api (to perform some action if a attribute changes) then is the set is successful, commit the attribute to the new value

dataplane/standalone/translator.h
DanG100: i believed there are, add todo to confirm

dataplane/standalone/translator.cc
wenovus: Okay that makes sense. Perhaps document this in attribute manager? I don't remember reading this.

Can `sai_object_id_t` be used as the ID instead of allocating? How does that ID get created?

dataplane/standalone/translator.cc
DanG100: i think it's different enough, i'm overloading the funcs in some cases, so it's actually calling different functions

dataplane/standalone/common.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "dataplane/standalone/common.h"
+
+extern "C" {
+#include "inc/sai.h"
+#include "meta/saimetadata.h"
+}
+
+sai_object_id_t AttributeManager::create(sai_object_type_t type,
+                                         sai_object_id_t switch_id) {
+  sai_object_id_t id = this->objects.size();
```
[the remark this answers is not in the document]
DanG100: i haven't seen any, added todo

dataplane/standalone/common.cc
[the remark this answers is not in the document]
DanG100: i've simplified this now: create either take an entry, in which case it serializes it and uses that as the id or it takes just a type and allocates a new id.

dataplane/standalone/translator.cc
DanG100: i've simplified this now: create either take an entry, in which case it serializes it and uses that as the id or it takes just a type and allocates a new id.

wenovus (approved):

PR #193, opened 2023-05-18
Handle port state in fwdport implementations

DanG100:


dataplane/forwarding/fwdport/ports/kernel.go
```diff
@@ -67,6 +71,7 @@ func (p *kernelPort) Cleanup() {
 	p.input.Cleanup()
 	p.output.Cleanup()
 	p.handle.Close()
+	close(p.linkDoneCh)
```
wenovus, line 74: close(p.linkUpdateCh) as well?

dataplane/handlers/interface.go
```diff
@@ -247,7 +251,16 @@ func (ni *Interface) reconcile(config *oc.Interface) {
 	if config.GetOrCreateSubinterface(0).Enabled != nil {
 		if state.GetOrCreateSubinterface(0).Enabled == nil || config.GetSubinterface(0).GetEnabled() != state.GetSubinterface(0).GetEnabled() {
 			log.V(1).Infof("setting interface %s enabled %t", tapName, config.GetSubinterface(0).GetEnabled())
-			if err := ni.ifaceMgr.SetState(tapName, config.GetSubinterface(0).GetEnabled()); err != nil {
+			state := fwdpb.PortState_PORT_STATE_DISABLED_DOWN
+			if config.GetSubinterface(0).GetEnabled() {
+				state = fwdpb.PortState_PORT_STATE_ENABLED_UP
+			}
+			_, err := ni.e.PortState(context.Background(), &fwdpb.PortStateRequest{
```
wenovus, line 258: Why is the return discarded?

wenovus (approved):

dataplane/forwarding/fwdport/ports/tap.go
```diff
@@ -129,21 +137,90 @@ func (p *tapPort) Actions(dir fwdpb.PortAction) fwdaction.Actions {
...
+	log.V(1).Infof("dataplane write port %q: admin state %v, oper state %v, setting to %v", name, adminState.String(), operState.String(), pi.AdminStatus.String())
+	if adminState == pi.GetAdminStatus() {
+		return reply, nil
+	}
+	if pi.AdminStatus == fwdpb.PortState_PORT_STATE_DISABLED_DOWN {
+		err = ifMgr.SetState(name, false)
+	} else if pi.AdminStatus == fwdpb.PortState_PORT_STATE_ENABLED_UP {
+		err = ifMgr.SetState(name, true)
+	}
+	return reply, err
```
wenovus, line 189: Here reply.Status is not updated after updating the state. Is this correct? If so can update the doc comment to explain what the behaviour of this function is.

dataplane/handlers/interface.go
DanG100: because the state update happens when a  notification is received, don't need to do twice

dataplane/forwarding/fwdport/ports/kernel.go
DanG100: sending on closed channel panics, so relying on the GC so close the channel once all goroutines are done with it. (not that it really matter in this case)

PR #194, opened 2023-05-19
Refactor gobgp tasks's reconciler

wenovus:
* Declare commonly used paths.
* Use updateAppliedState to allow concurrent threads to update the appliedState.
* Add path subscriptions for BGP policies (not implemented yet)

BGP integration tests pass


bgp/gobgp.go
```diff
@@ -62,15 +63,55 @@ type bgpDeclTask struct {
...
 		bgpServer:  server.NewBgpServer(),
 		listenPort: listenPort,
+
+		appliedState:         appliedState,
+		appliedBGP:           appliedBGP,
+		appliedRoutingPolicy: appliedRoutingPolicy,
+	}
+}
+
+func updateAppliedStateHelper[T ygot.GoStruct](yclient *ygnmi.Client, path ygnmi.SingletonQuery[T], appliedState T) {
```
DanG100, line 95: optional: inline this into updateAppliedState  so that someone doesn't call it instead of updateAppliedState. it's only 3 lines

DanG100 (approved):

bgp/gobgp.go
wenovus: I'm still just very slightly for keeping this since it might be applicable for other handlers

PR #197, opened 2023-05-23
Implement SAI port state notifications

DanG100:
* Change notifications a from a gRPC server implemented by the client to a streaming RPC
* Implement notification client in the switch


dataplane/forwarding/fwd.go
```diff
@@ -922,3 +877,23 @@ func (e *Server) Operation(stream fwdpb.Forwarding_OperationServer) error {
...
+
+// NotifySubscribe subscribes to notification for a forwarding context.
+// TODO: Only one notification client is supported per forwarding context.
+func (e *Server) NotifySubscribe(sub *fwdpb.NotifySubscribeRequest, srv fwdpb.Forwarding_NotifySubscribeServer) error {
+	eventCh := make(chan *fwdpb.EventDesc)
+	fn := func(ed *fwdpb.EventDesc) {
+		eventCh <- ed
+	}
+
+	if err := e.UpdateNotification(sub.GetContext(), fn, "callback"); err != nil {
```
wenovus, line 889: I guess the notification address is not useful anymore? Maybe a TODO comment to remove confusion?

dataplane/standalone/switch.cc
```diff
@@ -365,3 +374,40 @@ sai_status_t Switch::set_child_attr(sai_object_type_t type, std::string id,
                                     const sai_attribute_t *attr) {
   return this->apis[id]->set_attribute(attr);
 }
+
+void Switch::handle_notification() {
+  grpc::ClientContext ctx;
+  forwarding::NotifySubscribeRequest req;
+  req.mutable_context()->set_id("default");
```
wenovus, line 381: Should this be "DEFAULT"? Also is when the context is created is the name configurable? If so use a common variable?

wenovus (approved):

dataplane/standalone/switch.cc
DanG100: used a variable from go instead and changed name to avoid confusion with network-instance as it's not related

PR #198, opened 2023-05-26
Dataplane gRPC API for routes

DanG100:


proto/forwarding/forwarding_common.proto
```diff
@@ -168,7 +168,9 @@ enum PacketFieldNum {
   PACKET_FIELD_NUM_ARP_TMAC = 52;             // ARP target mac address.
   PACKET_FIELD_NUM_ARP_SMAC = 53;             // ARP source mac address.
   PACKET_FIELD_NUM_ARP_SPA = 54;              // ARP source protocol address.
-  PACKET_FIELD_NUM_NEXT_HOP_IP = 56; // Next hop ip (packet metadata).
+  PACKET_FIELD_NUM_NEXT_HOP_IP = 56; // Next hop ip (metadata).
+  PACKET_FIELD_NUM_NEXT_HOP_GROUP_ID = 57; // Next hop group id (metadata).
+  PACKET_FIELD_NUM_NEXT_HOP_ID = 58; // Next hop id (metadata).
```
wenovus, line 173: What's the difference between nexthop ID and IP? What is the usecase for ID?

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+
+import (
+	"encoding/binary"
+
+	fwdpb "github.com/openconfig/lemming/proto/forwarding"
+)
+
+type actionDescBuilder interface {
+	set(*fwdpb.ActionDesc)
+	atype() fwdpb.ActionType
```
wenovus, line 26: maybe it's better to be verbose here? actionType()?

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+var (
+	// Compile-time checks that builders implement actionDescBuilder interface.
+	_ actionDescBuilder = &LookupActionBuilder{}
+	_ actionDescBuilder = &TransmitActionBuilder{}
+	_ actionDescBuilder = &LookupActionBuilder{}
```
wenovus, line 33: First one is UpdateActionBuilder

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+// ActionBuilder is a builder for forward action types.
+type ActionBuilder struct {
+	adb actionDescBuilder
+}
+
+// Action returns a new action builder.
+func Action(adb actionDescBuilder) *ActionBuilder {
```
wenovus, line 42: I think it's clearer if this is called NewAction or actually removed because protobuf's builder API doesn't have this.

Also consider using the same API style as protobufs where you populate a builder struct -- this looks very similar to Ondatra's style but we may not necessarily need the verbosity of a fluent API.

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	ad.Action = upd
+}
+
+func (u *TransmitActionBuilder) atype() fwdpb.ActionType {
+	return fwdpb.ActionType_ACTION_TYPE_TRANSMIT
+}
+
+// TransmitActionBuilder is a builder for a transmit action.
+type LookupActionBuilder struct {
```
wenovus, line 169: update comment, same below

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+	tableID string
+}
+
+// TransmitAction returns a new update action builder.
+func LookupAction(tableID string) *LookupActionBuilder {
```
wenovus, line 174: NewLookupAction?

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+// Package fwdbuilder contains builders for varius proto types.
+package fwdbuilder
+
+import (
+	"encoding/binary"
+
+	fwdpb "github.com/openconfig/lemming/proto/forwarding"
+)
+
+type actionDescBuilder interface {
```
wenovus, line 24: Is this a builder type or a mutator type? Its methods seems to have the capability of mutating the given ActionDesc rather than returning a built object which is what I expected when I started reading the code.

dataplane/forwarding/fwdbuilder/action.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	ab.adb.set(a)
+	return a
+}
+
+// UpdateActionBuilder is a builder for an update action.
+type UpdateActionBuilder struct {
+	fieldIDNum fwdpb.PacketFieldNum
+	updateType fwdpb.UpdateType
+	fieldSrc   fwdpb.PacketFieldNum
```
wenovus, line 67: Is this the right type? What does this field represent?

proto/forwarding/forwarding_common.proto
DanG100: SAI tracks next hops by an id, it's easier to have that state management in the Lucius

dataplane/forwarding/fwdbuilder/action.go
DanG100: because the protobuf oneof generates an unexported interface, it's easier for this to mutate the parent on build

dataplane/forwarding/fwdbuilder/action.go
DanG100: yes, for COPY update type src header

dataplane/forwarding/fwdbuilder/action.go
DanG100: done

dataplane/forwarding/fwdbuilder/action.go
DanG100: done

dataplane/forwarding/fwdbuilder/action.go
DanG100: i prefer this approach https://github.com/kubernetes/client-go/blob/master/applyconfigurations/core/v1/pod.go

see next PR for how APU is used, it's a significant improvement

dataplane/forwarding/fwdbuilder/action.go
DanG100: following style of link in earlier, i think it's clear without the New prefix

dataplane/forwarding/fwdbuilder/action.go
wenovus: Thanks for the link. There the object is called AppliedConfiguration. Do you think it's clearer if the package is called `fwdconfig`?

dataplane/forwarding/fwdbuilder/action.go
wenovus: I think if the package name were `fwdconfig.LookupAction` then it is more clear. I think it's just the "builder" part that's confusing me since it's not the conventional builder type I was expecting.

dataplane/forwarding/fwdbuilder/action.go
DanG100: sgtm

dataplane/forwarding/fwdbuilder/action.go
DanG100: renamed

wenovus (approved):

PR #199, opened 2023-05-26
Add route to dataplane service

DanG100:
* Add the concept of next hop groups and next hop to route API
* Allow explicit or implicit creation of next hops and next hop groups


dataplane/internal/engine/engine.go
```diff
@@ -233,45 +282,249 @@ func (e *Engine) AddLayer3PuntRule(ctx context.Context, portName string, ip []by
...
+// ip addr bytes, ip mask bytes, is ipv4, vrf id, error.
+func prefixToPrimitives(prefix *dpb.RoutePrefix) ([]byte, []byte, bool, uint64, error) {
+	var ip []byte
+	var mask []byte
+	var isIPv4 bool
+	vrf := prefix.GetVrfId()
+
+	switch pre := prefix.GetPrefix().(type) {
+	case *dpb.RoutePrefix_Str:
+		_, ipNet, err := net.ParseCIDR(pre.Str)
```
wenovus, line 295: (no action required) I recommend checking out netip package, it was created because net was a bit hard to use.

dataplane/internal/engine/engine.go
```diff
@@ -233,45 +282,249 @@ func (e *Engine) AddLayer3PuntRule(ctx context.Context, portName string, ip []by
...
+	if _, err := e.Server.TableEntryAdd(ctx, entries); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// AddIPRoute adds a route to the FIB. It operates in two modes:
+// 1. Client-managed IDs: each next hop and next hop group must be created before adding to a route with user provided ids.
+// 2. Server-managed IDs: each next hop and next hop group must be specified with route. The server implicitly creates ids.
```
wenovus, line 493: Is the mode implicit via convention or enforced in some way?

dataplane/internal/engine/engine.go
```diff
@@ -233,45 +282,249 @@ func (e *Engine) AddLayer3PuntRule(ctx context.Context, portName string, ip []by
...
+		default:
+			return ip, mask, isIPv4, vrf, fmt.Errorf("invalid ip addr length")
+		}
+	default:
+		return ip, mask, isIPv4, vrf, fmt.Errorf("invalid prefix type")
+	}
+	return ip, mask, isIPv4, vrf, nil
+}
+
+// addNextHopList creates all the next hops from the message, then create a next hop group if there are multiple next hops.
```
wenovus, line 324: So technically each route never directly points to a nexthop ID, always NHG ID first. Is there a reason why you're providing direct access to NH IDs?

dataplane/internal/engine/engine.go
```diff
@@ -233,45 +282,249 @@ func (e *Engine) AddLayer3PuntRule(ctx context.Context, portName string, ip []by
...
+					},
+				},
+			}},
+		})
+	}
+	actions := []*fwdpb.ActionDesc{{
+		ActionType: fwdpb.ActionType_ACTION_TYPE_SELECT_ACTION_LIST,
+		Action: &fwdpb.ActionDesc_Select{
+			Select: &fwdpb.SelectActionListActionDesc{
+				SelectAlgorithm: fwdpb.SelectActionListActionDesc_SELECT_ALGORITHM_CRC32, // TODO: should algo + hash be configurable?
```
wenovus, line 382: I think the comment got lost in the edit -- annotate that this is for ECMP/WCMP.

dataplane/internal/engine/engine.go
```diff
@@ -233,45 +282,249 @@ func (e *Engine) AddLayer3PuntRule(ctx context.Context, portName string, ip []by
...
+		nextHopIP = nextHop.To4()
+		if nextHopIP == nil {
+			nextHopIP = nextHop.To16()
 		}
+	}
+	// Set the next hop IP in the packet's metadata.
+	nextHopAct := fwdbuilder.Action(fwdbuilder.UpdateAction(fwdpb.UpdateType_UPDATE_TYPE_SET, fwdpb.PacketFieldNum_PACKET_FIELD_NUM_NEXT_HOP_IP).WithValue(nextHopIP)).Build()
+	if nextHopIP == nil {
+		nextHopAct = fwdbuilder.Action(fwdbuilder.UpdateAction(fwdpb.UpdateType_UPDATE_TYPE_COPY, fwdpb.PacketFieldNum_PACKET_FIELD_NUM_NEXT_HOP_IP).WithFieldSrc(fwdpb.PacketFieldNum_PACKET_FIELD_NUM_IP_ADDR_DST)).Build()
+	}
```
wenovus, line 454: mention that if nexthop IP is not specified that the packet is treated as directly connected? This still doesn't quite make sense to me because it should be dropped or go to the default route right?

wenovus (approved):

proto/dataplane/dataplane.proto
```diff
@@ -54,6 +77,34 @@ message CreatePortRequest {
 message CreatePortResponse {
 }
 
+message IpMask {
+  bytes addr = 1;
+  bytes mask = 2;
+}
+
+enum PacketAction {
```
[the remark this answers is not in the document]
DanG100: used in route message

proto/dataplane/dataplane.proto
```diff
@@ -32,13 +32,36 @@ message NextHop {
...
 }
 
 
+
+message Route {
+  RoutePrefix prefix = 1;
+  PacketAction action = 2;
+  oneof hop  {
+    string port_id = 3; // Id of the output port.
+    uint64 next_hop_id = 4; // Id of the next hop
```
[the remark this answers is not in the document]
DanG100: no, nhg are only used if there are multiple nexts hops, otherwise the hashing is unnecessary

dataplane/internal/engine/engine.go
DanG100: incorrect, routes can point directly to next hops

dataplane/internal/engine/engine.go
DanG100: agreed added todo remove this (was old behavior kept for compat with how sysrib calls this API)

dataplane/internal/engine/engine.go
DanG100: added todo

dataplane/handlers/routes.go
```diff
@@ -66,30 +65,18 @@ func (r *route) start(ctx context.Context, client *ygnmi.Client) error {
...
 		if !present {
-			if err := r.e.DeleteIPRoute(ctx, isIPv4, ipNet.IP, ipNet.Mask, vrf); err != nil {
+			if _, err := r.e.RemoveIPRoute(ctx, &dpb.RemoveIPRouteRequest{Prefix: &dpb.RoutePrefix{Prefix: &dpb.RoutePrefix_Str{Str: prefix}, VrfId: vrf}}); err != nil {
 				log.Warningf("failed to delete route: %v", err)
 				return ygnmi.Continue
 			}
 			return ygnmi.Continue
 		}
-		if len(route.NextHops) == 0 {
+		if route.GetNextHops() == nil || len(route.GetNextHops().GetHops()) == 0 {
```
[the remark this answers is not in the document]
DanG100: yes, fixed

dataplane/internal/engine/engine.go
DanG100: it's easier to get the mask bytes with net package, but yeah otherwise netip is nice

PR #195, opened 2023-05-19
Support prefix-list BGP policies

wenovus:
* Convert OC policies to GoBGP policies in configuration in the reconciler.
* Add BGP policy test spec. This is adapted from an existing one.
* Add prefix-list policy test.

PR #200, opened 2023-05-30
Basic sai route handling

DanG100:


wenovus:
`update is stale`: I've seen this error before at https://github.com/openconfig/lemming/pull/195#issuecomment-1558150260, not sure what's causing it and I've only seen it happen on GitHub actions.

I can take another look afterwards. I don't think it's related to your change.

dataplane/standalone/route.cc
```diff
@@ -14,13 +14,107 @@
...
+    switch (obj_type) {
+      case SAI_OBJECT_TYPE_NEXT_HOP:
+        req.mutable_route()->set_next_hop_id(oid);
+        break;
+      case SAI_OBJECT_TYPE_NEXT_HOP_GROUP:
+        req.mutable_route()->set_next_hop_group_id(oid);
+        break;
+      case SAI_OBJECT_TYPE_ROUTER_INTERFACE:
+        // TODO(dgrau): Support router interfaces.
+        break;
```
wenovus, line 98: If it's unsupported should we return an error status or is that only in strict mode?

wenovus (approved):

PR #201, opened 2023-06-01
Implement sai next hops

DanG100:


dataplane/standalone/route.h
```diff
@@ -41,4 +41,43 @@ class Route : public APIBase {
   sai_status_t set_attribute(_In_ const sai_attribute_t* attr);
 };
 
+class NextHop : public APIBase {
```
wenovus, line 44: NextHop and NHGs are in separate header files in the SAI project. Would be better if they're split out in the same organization so it's more obvious how to reference between the two?

dataplane/internal/engine/engine.go
```diff
@@ -353,11 +355,21 @@ func (e *Engine) addNextHopList(ctx context.Context, nhg *dpb.NextHopList) ([]*f
...
-func (e *Engine) addNextHopGroupIDList(ctx context.Context, id uint64, nhg *dpb.NextHopIDList) error {
+func (e *Engine) addNextHopGroupIDList(ctx context.Context, id uint64, nhg *dpb.NextHopIDList, mode dpb.GroupUpdateMode) error {
+	e.nextHopGroupsMu.Lock()
+	defer e.nextHopGroupsMu.Unlock()
+
+	hops := nhg.GetHops()
+	weights := nhg.GetWeights()
+	if mode == dpb.GroupUpdateMode_GROUP_UPDATE_MODE_APPEND {
+		hops = append(hops, e.nextHopGroups[id].Hops...)
+		weights = append(weights, e.nextHopGroups[id].Weights...)
```
wenovus, line 366: nit: this seems to be prepending instead of appending the new hops which is initially confusing.

dataplane/internal/engine/engine.go
```diff
@@ -430,9 +442,24 @@ func (e *Engine) addNextHopGroupIDList(ctx context.Context, id uint64, nhg *dpb.
 			Actions: actions,
 		}},
 	}
+
+	if mode == dpb.GroupUpdateMode_GROUP_UPDATE_MODE_APPEND || mode == dpb.GroupUpdateMode_GROUP_UPDATE_MODE_REPLACE {
```
wenovus, line 446: handle all enum values with a default case. If new one is added and this is not updated then UNSPECIFIED's behaviour will be used instead of throwing an error.

dataplane/internal/engine/engine.go
```diff
@@ -516,7 +545,7 @@ func (e *Engine) actionsFromRoute(ctx context.Context, route *dpb.Route) ([]*fwd
 		}
 	case *dpb.Route_NextHops:
 		var err error
-		actions, err = e.addNextHopList(ctx, hop.NextHops)
+		actions, err = e.addNextHopList(ctx, hop.NextHops, dpb.GroupUpdateMode_GROUP_UPDATE_MODE_UNSPECIFIED)
```
wenovus, line 548: Would adding a new enum value be better than UNSPECIFIED? Can use that as default behaviour.

dataplane/standalone/route.h
DanG100: done

dataplane/internal/engine/engine.go
DanG100: changed

dataplane/internal/engine/engine.go
DanG100: done

dataplane/internal/engine/engine.go
DanG100: changed unspecified to ERROR ON CONFLICT, by making id 0, if unset it's the default bahvior

wenovus (approved):

PR #202, opened 2023-06-12
Update ondatra + related deps

DanG100:


patches/gnoi.patch
wenovus, line 1: Why did we need these patches again?

wenovus (approved):

patches/gnoi.patch
DanG100: i have a pending PR for gnoi repo with these changes. it's minor tweaks to the build rules so they compatible with lemming

PR #205, opened 2023-06-14
Add model to topos and remove execs of kne_cli

DanG100:
* "LEMMING" model to topologies
* change lemming bind to kne as library instead of cli
* enable incompatible strict action env to improve bazel caching

### DIFF
--- a/dataplane/internal/engine/engine.go
+++ b/dataplane/internal/engine/engine.go
@@ -62,9 +62,10 @@ type Engine struct {
 // New creates a new engine and sets up the forwarding tables.
 func New(ctx context.Context) (*Engine, error) {
 	e := &Engine{
-		id:      "lucius",
-		Server:  forwarding.New("engine"),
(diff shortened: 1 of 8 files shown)


dataplane/internal/engine/engine.go
```diff
@@ -62,9 +62,10 @@ type Engine struct {
 // New creates a new engine and sets up the forwarding tables.
 func New(ctx context.Context) (*Engine, error) {
 	e := &Engine{
-		id:      "lucius",
-		Server:  forwarding.New("engine"),
-		idToNID: map[string]uint64{},
+		id:            "lucius",
+		Server:        forwarding.New("engine"),
+		idToNID:       map[string]uint64{},
+		nextHopGroups: map[uint64]*dpb.NextHopIDList{},
```
wenovus, line 68: Was this intended to be part of this PR?

PR #206, opened 2023-06-15
Add support for loopback and cpu port

DanG100:
* add cpu port type. It expected that one existing on switch creating
* add router interface support for loopback ports. I haven't router interface called anywhere else


dataplane/internal/engine/engine.go
```diff
@@ -232,6 +232,20 @@ func (e *Engine) CreatePort(ctx context.Context, req *dpb.CreatePortRequest) (*d
 		err = e.CreateExternalPort(ctx, req.GetId(), req.GetKernelDev())
 	case fwdpb.PortType_PORT_TYPE_TAP:
 		err = e.CreateLocalPort(ctx, req.GetId(), req.GetKernelDev(), req.GetExternalPort())
+	case fwdpb.PortType_PORT_TYPE_CPU_PORT:
+		req := &fwdpb.PortCreateRequest{
+			ContextId: &fwdpb.ContextId{Id: e.id},
+			Port: &fwdpb.PortDesc{
+				PortId:   &fwdpb.PortId{ObjectId: &fwdpb.ObjectId{Id: req.GetId()}},
+				PortType: req.GetType(),
```
wenovus, line 240: nit: explicitly use `fwdpb.PortType_PORT_TYPE_CPU_PORT`

dataplane/standalone/switch.cc
```diff
@@ -77,6 +77,14 @@ sai_status_t Switch::create(_In_ uint32_t attr_count,
...
 
   auto portOid = this->attrMgr->create(SAI_OBJECT_TYPE_PORT, this->id);
+  std::vector<sai_attribute_t> portAttrs;
+  portAttrs.push_back({
+      .id = SAI_PORT_ATTR_TYPE,
+      .value = {.s32 = SAI_PORT_TYPE_CPU},
+  });
+  this->create_child(SAI_OBJECT_TYPE_PORT, portOid, portAttrs.size(),
+                     portAttrs.data());
+
```
wenovus, line 87: Shall we add a comment saying that SAI expects us to always have a CPU port?

dataplane/standalone/router_interface.cc
```diff
@@ -14,10 +14,41 @@
...
+        type = static_cast<sai_router_interface_type_t>(attr.value.s32);
+        break;
+    }
+  }
+  if (type == SAI_ROUTER_INTERFACE_TYPE_LOOPBACK) {
+    grpc::ClientContext context;
+    lemming::dataplane::CreatePortRequest req;
+    lemming::dataplane::CreatePortResponse resp;
+    req.set_id(this->id);
+    req.set_type(forwarding::PORT_TYPE_KERNEL);
```
wenovus, line 44: Is this the right type? How does a loopback interface work on lemming?

dataplane/standalone/port.cc
```diff
@@ -41,17 +45,15 @@ sai_status_t Port::create(_In_ uint32_t attr_count,
...
-    lemming::dataplane::CreatePortResponse resp;
-    req.set_id(this->id);
-    req.set_type(forwarding::PORT_TYPE_KERNEL);
-    req.set_kernel_dev(name);
+  grpc::ClientContext context;
+  lemming::dataplane::CreatePortRequest req;
+  lemming::dataplane::CreatePortResponse resp;
+  req.set_id(this->id);
+  if (type == SAI_PORT_TYPE_CPU) {
+    req.set_type(forwarding::PORT_TYPE_CPU_PORT);
```
wenovus, line 56: Looks like the if only needs to be on setting the req fields rather than the dataplane call handling? Can we restructure this so that the handling code is not duplicated?

dataplane/standalone/router_interface.cc
DanG100: for now using the kernel's loopback interface (lo)

wenovus (approved):

PR #208, opened 2023-06-17
Configure default neighbour policies to be ACCEPT_ROUTE.

wenovus:
This fixes the failing integration tests.

[OpenConfig makes the default REJECT_ROUTE](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-bgp-neighbors-neighbor-apply-policy-config-default-import-policy).


wenovus:
> Looks integration tests passed, looks like test failed on cleanup.
> 
> Do you know why the //bgp/local-test is failing?

I've seen this before. It's an error I've only seen on GitHub actions. Adding more debugging info.

DanG100 (approved):

wenovus:
I will keep the debugging log for when the error happens again.

PR #209, opened 2023-06-20
Improve presubmit speed by using a bazel cache

DanG100:


cloudbuild/presubmit.yaml
```diff
@@ -19,7 +19,7 @@ steps:
       - USERNAME=user
       - SSH_ARGS=--internal-ip --ssh-key-expire-after=1d
       - INSTANCE_NAME=kne-presubmit-$BUILD_ID
-      - INSTANCE_ARGS=--network cloudbuild-workers --image-project gep-kne --image-family kne --machine-type e2-standard-16 --preemptible
+      - INSTANCE_ARGS=--network cloudbuild-workers --image-project gep-kne --image-family kne --machine-type e2-standard-16 --preemptible --scopes=cloud-platform
```
wenovus, line 22: What does this flag do?

wenovus (approved):

cloudbuild/presubmit.yaml
DanG100: it grants the VM access to the various gcloud APIs. for some reason VMs need a scope and the service account they use needs access to resource

PR #211, opened 2023-06-22
Output logging for CI tests.

wenovus:
Useful for debugging presubmit failures


DanG100:
It looks this is causing new failures because the flags aren't recognized

wenovus:
> It looks this is causing new failures because the flags aren't recognized

I'll just use this PR to debug instead of trying to merge it.

PR #213, opened 2023-06-26
Run set of featureprofiles tests as part of presubmit

DanG100:


cloudbuild/fp-test.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+
+git clone https://github.com/openconfig/featureprofiles.git
+cd ... || exit
+
+function metadata_kne_topology() {
```
alexmasi, line 7: this can be simplified to always use the openconfig/lemming/topology.textproto as the resources differences between topos is negligible (not that this approach is wrong)

DanG100: thanks, done

alexmasi (approved):

PR #224, opened 2023-08-10
SAI proto gen part 3

DanG100:
* Generate all types that aren't attributes or message in common file
* Add entry types and fix some of the hand crafted types
* Correctly handle all list types.


wenovus (commented):
Should we add a very simple test here now that the code is more mature?

DanG100:
> Should we add a very simple test here now that the code is more mature?

I'd like to do this after the implementing the streaming RPCs.

wenovus (approved):

PR #229, opened 2023-08-14
Make lemming buildable on non-Linux envs

DanG100:
This is just a quick fix, I'd like to take a deeper look on the possibility of still being to run the dataplane on not linux


dataplane/reconcilers_nonlinux.go
```diff
@@ -0,0 +1,28 @@
...
+package dataplane
+
+import (
+	"github.com/openconfig/lemming/dataplane/handlers"
+	"github.com/openconfig/lemming/dataplane/internal/engine"
+	"github.com/openconfig/lemming/gnmi/reconciler"
+)
+
+func getReconcilers(e *engine.Engine) []reconciler.Reconciler {
+	return []reconciler.Reconciler{handlers.NewRoute(e)}
```
wenovus, line 27: On a non-linux machine would this mean that the dataplane receives routes, programs tables, but doesn't actually do anything else (e.g. react to interface config and reflect their states). If so should there be another such build constraint for lemming.New such that we error out if dataplane is turned on? https://github.com/openconfig/lemming/commit/e0ca9054bc2ba84782a0f41725ec8ebad17b8000

wenovus (commented):
Thanks this looks neat that we're able to separate out the linux-only interface manager such that it's not compiled unless we're on a linux machine.

dataplane/reconcilers_nonlinux.go
DanG100: I've implemented this a runtime check, to reduce the number of _linux files we have

lemming.go
```diff
@@ -152,6 +154,10 @@ func New(targetName, zapiURL string, opts ...Option) (*Device, error) {
 	var dplane *dataplane.Dataplane
 	var recs []reconciler.Reconciler
 
+	if viper.GetBool("enable_dataplane") && runtime.GOOS != "linux" {
+		return nil, fmt.Errorf("dataplane only supported on linux, GOOS is %s", runtime.GOOS)
+	}
+
 	if viper.GetBool("enable_dataplane") {
```
wenovus, line 161: Do the check in the existing if?

wenovus (approved):

PR #230, opened 2023-08-15
Don't use preemptible VMs for presubmit + other fixes

DanG100:
* Using not preemptible should help reduce flakiness
* The local test is unstable to github actions so run it on cloudbuild
* I'm not that `make load` was working as expected

### DIFF
--- a/bgp/tests/local_tests/BUILD
+++ b/bgp/tests/local_tests/BUILD
@@ -2,7 +2,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_test")
 
 go_test(
     name = "local_tests_test",
-    size = "medium",
+    size = "large",
     srcs = [
         "policy_test.go",
         "prefix_set_test.go",
--- a/BUILD
+++ b/BUILD
@@ -36,6 +36,7 @@ go_library(
 
 go_test(
     name = "lemming_test",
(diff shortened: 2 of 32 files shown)


alexmasi (approved):

bgp/tests/local_tests/BUILD
```diff
@@ -2,7 +2,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_test")
 
 go_test(
     name = "local_tests_test",
-    size = "medium",
+    size = "large",
```
wenovus, line 5: I think local_tests still counts as a medium since it runs for a few minutes right now. I also think this should be run as part of GitHub Actions rather than on cloudbuild. Perhaps add another workflow job that runs the medium jobs?

DanG100: the local test has been flaky on the github action runner, i want to see if solves the issue https://github.com/openconfig/lemming/actions/runs/5860193044/job/15887753577
https://github.com/openconfig/lemming/actions/runs/5686228980/job/15412660877

wenovus: Ok LGTM

wenovus (approved):

PR #235, opened 2023-08-21
Fix <PolicyTest>/installPolicyAfterRoutes

wenovus:
Need to check routes after reset.

Cannot use regular `Await` because need t.Fatal to be called in main test goroutine.


DanG100 (approved):

DanG100:
please fix lint issue

PR #237, opened 2023-08-22
Refactor Policy Tests

wenovus:
* Expanded generic policy_test topology to accommodate different tests.
* Factored out route propagation checking logic, which now can also check a new `NOT_PREFERRED` state.
* lemming device creation is done more automatically, with each listening only on its own address instead of on all addresses.
* Parallelize session establishment.
* community_set_test.go will be deleted in the next PR so can skip reviewing that.
* Install filtering policies on the import side due to GoBGP not supporting route withdrawals when soft-resetting on the export side.


wenovus:
Converting to a draft for now to see if the test time can be reduced, and potentially remove test from running on GitHub.

wenovus:
The "update is stale" error is appearing now on GCB as well. I will investigate this later. It's clear it's not just happening on GitHub.

DanG100 (approved):

PR #238, opened 2023-08-23
Support set-local-pref and set-med

wenovus:
Refactored community-set test for testing other attributes:
* There is now only one policy-setting phase for simplification.
* Each route has its own policy statement switch case block for creating related policies.


wenovus:
Verified integration tests locally using `go test integration_tests/...`

bgp/tests/local_tests/attributes_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+				var installDut5Stmt bool
+				dut5Stmt := &oc.RoutingPolicy_PolicyDefinition_Statement{Name: ygot.String(route + "-setattr-policy-dut5")}
+				dut5Stmt.GetOrCreateConditions().GetOrCreateMatchPrefixSet().SetPrefixSet(prefixSetName)
+				dut5Stmt.GetOrCreateConditions().GetOrCreateMatchPrefixSet().SetMatchSetOptions(oc.RoutingPolicy_MatchSetOptionsRestrictedType_ANY)
+
+				var installRejectStmt bool
+				rejectStmt := &oc.RoutingPolicy_PolicyDefinition_Statement{Name: ygot.String(route + "-reject-policy")}
+
+				switch i {
```
DanG100, line 140: this isn't very clear. i don't see why i == 1 means set communites

wenovus: Added a comment. Each index refers to an input test route, and each one has its corresponding policies that applies to it. Each one is essentially a subtest that re-uses the topology.

DanG100 (approved):

PR #242, opened 2023-08-25
Support AS Path Prepend

wenovus:


bgp/tests/local_tests/attributes_test.go
```diff
@@ -53,6 +53,8 @@ func TestAttributes(t *testing.T) {
 		"10.11.0.0/16",
 		"10.12.0.0/16",
 		"10.13.0.0/16",
+		"10.14.0.0/16",
```
DanG100, line 56: consider refactoring this test, it seems its testing to many and it's not clear to me what's going on here

PR #245, opened 2023-08-25
Generated proto and create calls

DanG100:


wenovus (commented):
Just ignore the CPP style? Also can you check why the check generate code checker is failing?

DanG100:
sorry, this was meant as a draft to make reviewing the generation logic easier. this isn't ready to be checked in yet

PR #251, opened 2023-09-01
Add IPv4 prefix set integration test

wenovus:
* policytest is adapted from local_test's policy_test.go and is suitable for being used in Ondatra tests instead of locally.

Previous PR: https://github.com/openconfig/lemming/pull/253


DanG100:
What's the long term plan for this? It seems like there's a lot duplication (code and test behavior) of the local test.

wenovus:
> What's the long term plan for this? It seems like there's a lot duplication (code and test behavior) of the local test.

I agree there is a lot of duplication. I think an Ondatra binding or generics/interfaces may be the solution to deal with the different Device/Port/ygnmi API type differences, as otherwise the two are identical.

My judgement is that local_tests will be useful as a lightweight way to test and debug BGP policy development, so once the Ondatra version of the test is stable, I'll look at an Ondatra bind for local lemmings.

PR #265, opened 2023-09-12
CreateSwitch and a few bug fixes

DanG100:
* fixed populating fields
* a createswitch

### DIFF
--- a/dataplane/standalone/lucius/lucius.go
+++ b/dataplane/standalone/lucius/lucius.go
@@ -18,16 +18,17 @@ import "C"
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"net"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
-
 	"github.com/openconfig/lemming/dataplane/internal/engine"
 	"github.com/openconfig/lemming/dataplane/standalone/saiserver"
+	"github.com/openconfig/lemming/dataplane/standalone/saiserver/attrmgr"
 
 	log "github.com/golang/glog"
 
@@ -60,6 +61,7 @@ func getLogger() logging.Logger {
 //export initialize
 func initialize(port int) {
 	log.Info("lucius initialized")
+
 	e, err := engine.New(context.Background())
 	if err != nil {
 		log.Fatalf("failed create engine: %v", err)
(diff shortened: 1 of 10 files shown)


dataplane/standalone/lucius/lucius.go
```diff
@@ -87,3 +91,8 @@ func initialize(port int) {
 
 func main() {
 }
+
+func init() {
+	// TODO: Figure a better way to config this when used standalone.
+	flag.Set("log_dir", "/var/log/syncd")
```
bormanp, line 97: Why is this being done?  It is possible to find which flags were actually set by using flag.Visit.

dataplane/standalone/saiserver/attrmgr/attrmgr.go
```diff
@@ -63,6 +68,27 @@ func (mgr *AttrMgr) set(id string, attr int32, val protoreflect.Value) {
 
 const protoNS = "lemming.dataplane.sai"
 
+// InvokeAndSave calls the RPC method and saves the attributes in the request.
```
bormanp, line 71: The description should indicate what it returns.

dataplane/standalone/saiserver/attrmgr/attrmgr.go
```diff
@@ -63,6 +68,27 @@ func (mgr *AttrMgr) set(id string, attr int32, val protoreflect.Value) {
...
+// This is the same behavior as the Interceptor, except for invoking server methods directly (not using gRPC).
+func InvokeAndSave[T proto.Message, S proto.Message](ctx context.Context, mgr *AttrMgr, rpc func(context.Context, T) (S, error), req T) (S, error) {
```
bormanp, line 73: Why are you using generics here?  Since T and S are both proto.Message and proto.Message is an interface can't you just use them directly?

dataplane/standalone/saiserver/switch.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+	hashResp, err := attrmgr.InvokeAndSave(ctx, sw.mgr, sw.hash.CreateHash, &saipb.CreateHashRequest{
+		Switch: swID,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// These values are mostly meaningless, but clients expect these to be set.
```
bormanp, line 110: Please include some reference to where these magic numbers came from.

dataplane/standalone/saiserver/attrmgr/attrmgr.go
DanG100: By using generics you can do
`InvokeAndSave(ctx,mgr, port.CreatePort, portReq)`
otherwise you'd have to do
```
InvokeAndSave(ctx, mgr, func(ctx context.Context, req proto.Message) (proto, error) {
   return port.CreatePort(ctx,req)
}, portReq)
```

dataplane/standalone/lucius/lucius.go
DanG100: this just temporary until we run the dataplane in its own container (and can set the args)

dataplane/standalone/saiserver/attrmgr/attrmgr.go
DanG100: done

dataplane/standalone/saiserver/switch.go
DanG100: done

bormanp (approved):

PR #266, opened 2023-09-13
Implement sai port APIs

DanG100:


dataplane/standalone/saiserver/ports.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		if req.GetAdminState() {
+			stateReq.Operation.AdminStatus = fwdpb.PortState_PORT_STATE_ENABLED_UP
+		}
+		_, err := port.dataplane.PortState(ctx, stateReq)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nil, nil
```
alexmasi, line 124: should this return

saipb.SetPortAttributeResponse{}, nil

? I know its basically the same thing

dataplane/standalone/saiserver/switch_test.go
```diff
@@ -102,15 +102,15 @@ func TestCreateSwitch(t *testing.T) {
 	}
 }
 
-func newTestSwitch(t testing.TB) (saipb.SwitchClient, *attrmgr.AttrMgr, func()) {
+func newTestServer(t testing.TB, newSrvFn func(mgr *attrmgr.AttrMgr, srv *grpc.Server)) (grpc.ClientConnInterface, *attrmgr.AttrMgr, func()) {
```
alexmasi, line 105: could newSrvFn be an option in case a test wants to use a default fn like before?

alexmasi (approved):

dataplane/standalone/saiserver/ports.go
DanG100: done

dataplane/standalone/saiserver/switch_test.go
DanG100: made the func optional by passing nil

PR #278, opened 2023-09-27
Support empty vs nil lists and error for unset attrs

DanG100:
* Return error is attribute isn't set
* Support empty vs nil lists
  * Empty list: value is set  and is empty
  * Nil: value is unset

### DIFF
--- a/dataplane/standalone/saiserver/attrmgr/attrmgr.go
+++ b/dataplane/standalone/saiserver/attrmgr/attrmgr.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	log "github.com/golang/glog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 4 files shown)


dataplane/standalone/saiserver/attrmgr/attrmgr.go
```diff
@@ -173,9 +175,14 @@ func (mgr *AttrMgr) PopulateAttributes(req, resp proto.Message) error {
 		enumVal := reqList.Get(i).Enum()
 		val, ok := mgr.attrs[id][int32(enumVal)]
 		if !ok {
-			continue
+			// log.Errorf("requested attribute not set: id %v, obj type %v, enum %v", id, mgr.GetType(id), reqList.Get(i))
```
wenovus, line 178: stale debug log?

wenovus (approved):
What's the purpose of empty vs. nil? Can you document that?

DanG100:
> What's the purpose of empty vs. nil? Can you document that?

add comment

PR #279, opened 2023-09-28
Increase Await timeout for BGP policy test

wenovus:
Recent failures are at BGP session establishment, which is the place of highest CPU usage.


DanG100 (approved):

wenovus:
Merging now since the build failure is unrelated.

PR #281, opened 2023-09-28
Cache build and test separately

DanG100:
Since not all build targets have tests, the build cache is more complete than the test cache


wenovus:
It looks like gNMI was not able to get the OTG's interface mac-addresses. I'm re-running the test to see if it fixes it.

The

PR #283, opened 2023-10-02
Implement dataplane reset

DanG100:
* Implement timeouts of packet io, so they close correctly
* Add reset to engine and sai API


dataplane/standalone/proto/common.proto
```diff
@@ -260,9 +260,19 @@ message ObjectTypeQueryResponse {
...
 }
 
+message InitializeRequest {}
+
+message InitializeResponse {}
+
+message UninitializeRequest {}
+
+message UninitializeResponse {}
+
```
wenovus, line 270: Are these being used?

wenovus (approved):
How will you test this change?

DanG100:
> How will you test this change?

I have task to for an integration test for this to do later

dataplane/standalone/proto/common.proto
DanG100: not currently. for reset need to figure out if we need to start and stop in a specific order.

PR #291, opened 2023-10-09
Don't require Target when doing gNMI requests

### LINKED ISSUE #263: Subscribe() requires a target while the spec says it shouldn't
Hey!

Subscribe() [requires a target](https://github.com/openconfig/gnmi/blob/5473f2ef722ee45c3f26eee3f4a44a7d827e3575/subscribe/subscribe.go#L270), while the spec says it should not. In [2.2.2.1](https://www.openconfig.net/docs/gnmi/gnmi-specification/#2221-path-target):

>  This field is optional for clients.

Commercial implementations do not require a target. Currently, `Subscribe()` implementation is delegated to `openconfig/gnmi` which requires a target. See https://github.com/openconfig/gnmi/issues/151.

wenovus:
* Auto-populate Target when doing gNMI requests
* Auto-remove Target when responding to gNMI SubscribeRequest that didn't have prefix populated: https://www.openconfig.net/docs/gnmi/gnmi-specification/#2221-path-target

Resolves #263

### DIFF
--- a/gnmi/BUILD
+++ b/gnmi/BUILD
@@ -27,6 +27,7 @@ go_library(
         "@org_golang_google_grpc//peer",
         "@org_golang_google_grpc//status",
         "@org_golang_google_protobuf//encoding/prototext",
+        "@org_golang_google_protobuf//proto",
         "@org_golang_x_exp//slices",
     ],
 )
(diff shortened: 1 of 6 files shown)


DanG100:
Is there a TODO to remove all the places in the lemming that accept a target?

DanG100 (approved):

wenovus:
> Is there a TODO to remove all the places in the lemming that set a target?

Thanks for catching this. There are indeed a lot of places the code that now unnecessarily take `target` as an argument or otherwise assume that gNMI requires populating `target`. I'll create a task for clean-up.

PR #293, opened 2023-10-10
Initial support of acl RPCs

DanG100:
These tables aren't used anywhere yet, coming in future PR
Note: i will add unit tests one the tables are integrated in the forwarding pipeline as there may be some impl changes

### DIFF
--- a/dataplane/standalone/saiserver/acl.go
+++ b/dataplane/standalone/saiserver/acl.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:revive // Acl is incorrect, but it comes from generated code.
 package saiserver
 
 import (
@@ -123,8 +124,7 @@ func (a *acl) CreateAclTableGroupMember(_ context.Context, req *saipb.CreateAclT
 }
 
 // CreateAclTable is noop as the table is already created in the group.
-// TODO: Do we need to support tables that aren't in groups.
-func (a *acl) CreateAclTable(_ context.Context, req *saipb.CreateAclTableRequest) (*saipb.CreateAclTableResponse, error) {
+func (a *acl) CreateAclTable(context.Context, *saipb.CreateAclTableRequest) (*saipb.CreateAclTableResponse, error) {
 	id := a.mgr.NextID()
 	return &saipb.CreateAclTableResponse{Oid: id}, nil
 }
--- a/dataplane/standalone/saiserver/acl.go
+++ b/dataplane/standalone/saiserver/acl.go
@@ -125,7 +125,7 @@ func (a *acl) CreateAclTable(_ context.Context, req *saipb.CreateAclTableRequest
 }
 
 // CreateAclEntry adds an entry in the a bank.
-func (a *acl) CreateAclEntry(_ context.Context, req *saipb.CreateAclEntryRequest) (*saipb.CreateAclEntryResponse, error) {
+func (a *acl) CreateAclEntry(ctx context.Context, req *saipb.CreateAclEntryRequest) (*saipb.CreateAclEntryResponse, error) {
 	id := a.mgr.NextID()
 	gb, ok := a.tableToBank[req.GetTableId()]
 	if !ok {
--- a/dataplane/standalone/saiserver/acl.go
+++ b/dataplane/standalone/saiserver/acl.go
@@ -101,15 +101,15 @@ func (a *acl) CreateAclTableGroup(ctx context.Context, req *saipb.CreateAclTable
 }
 
 // CreateAclTableGroupMember stores the acl table id to its corresponding lucius flow table and bank.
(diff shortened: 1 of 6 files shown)


dataplane/standalone/saiserver/acl.go
```diff
@@ -0,0 +1,167 @@
...
+	ID() string
+	TableCreate(context.Context, *fwdpb.TableCreateRequest) (*fwdpb.TableCreateReply, error)
+	TableEntryAdd(context.Context, *fwdpb.TableEntryAddRequest) (*fwdpb.TableEntryAddReply, error)
+}
+
+type groupBank struct {
+	groupID  string
+	bank     int
+	memberID uint64
+}
```
wenovus, line 42: What's a group bank?

dataplane/standalone/saiserver/acl.go
```diff
@@ -0,0 +1,167 @@
...
+
+// CreateAclTableGroupMember stores the acl table id to its corresponding lucius flow table and bank.
+func (a *acl) CreateAclTableGroupMember(_ context.Context, req *saipb.CreateAclTableGroupMemberRequest) (*saipb.CreateAclTableGroupMemberResponse, error) {
+	groupID := req.GetAclTableGroupId()
+	tableID := req.GetAclTableId()
+	memberID := a.mgr.NextID()
+
+	bank := a.groupNextFreeBank[groupID]
+	a.groupNextFreeBank[groupID] = bank + 1
```
wenovus, line 110: Use atomic library for concurrency?

dataplane/standalone/saiserver/acl.go
```diff
@@ -0,0 +1,167 @@
...
+type aclDataplaneAPI interface {
+	ID() string
+	TableCreate(context.Context, *fwdpb.TableCreateRequest) (*fwdpb.TableCreateReply, error)
+	TableEntryAdd(context.Context, *fwdpb.TableEntryAddRequest) (*fwdpb.TableEntryAddReply, error)
+}
+
+type groupBank struct {
+	groupID  string
+	bank     int
+	memberID uint64
```
wenovus, line 41: What's the purpose of memberID?

dataplane/standalone/saiserver/acl.go
```diff
@@ -0,0 +1,167 @@
...
+// CreateAclTable is noop as the table is already created in the group.
+// TODO: Do we need to support tables that aren't in groups.
+func (a *acl) CreateAclTable(_ context.Context, req *saipb.CreateAclTableRequest) (*saipb.CreateAclTableResponse, error) {
+	id := a.mgr.NextID()
+	return &saipb.CreateAclTableResponse{Oid: id}, nil
+}
+
+// CreateAclEntry adds an entry in the a bank.
+func (a *acl) CreateAclEntry(ctx context.Context, req *saipb.CreateAclEntryRequest) (*saipb.CreateAclEntryResponse, error) {
+	id := a.mgr.NextID()
```
wenovus, line 129: So NextID() is also called here in addition to CreateAclTableGroupMember, is that right? I'm not really understanding the code but wanted to point that out.

dataplane/standalone/saiserver/acl.go
```diff
@@ -0,0 +1,167 @@
...
+					Id:       uint32(id),
+					Bank:     uint32(gb.bank),
+				},
+			},
+		},
+	}
+	switch {
+	case req.GetFieldDstIp() != nil:
+		aReq.EntryDesc.GetFlow().Fields = append(aReq.EntryDesc.GetFlow().Fields, &fwdpb.PacketFieldMaskedBytes{
+			FieldId: &fwdpb.PacketFieldId{Field: &fwdpb.PacketField{FieldNum: fwdpb.PacketFieldNum_PACKET_FIELD_NUM_IP_ADDR_DST}},
```
wenovus, line 151: Just for my understanding, is the IP mask identifier used to identify which flows fall under the table entry? So if the IP mask is 10/8, then 10.10.10.10 would be affected by this table entry?

wenovus (commented):
A few questions so I can get a high-level understanding.

dataplane/standalone/saiserver/acl.go
DanG100: yes

dataplane/standalone/saiserver/acl.go
DanG100: yes, every object created needs an id

dataplane/standalone/saiserver/acl.go
DanG100: need to save the  id of the member in case it gets deleted later

dataplane/standalone/saiserver/acl.go
DanG100: added mu

wenovus (approved):
Approved after lints are fixed.

PR #296, opened 2023-10-12
Make sysrib tests end-to-end.

wenovus:
Test final output (i.e. internal dataplane gNMI path) as opposed to testing how `ProgramRoute` is called.


sysrib/zapi_test.go
```diff
@@ -173,6 +174,35 @@ func testRouteAdd(t *testing.T) {
 	}
 }
 
+// TODO(wenbli): Remove
+type FakeDataplane struct {
```
DanG100, line 178: is this used somewhere? does it need to be exported?

DanG100 (approved):

sysrib/zapi_test.go
wenovus: Yes it's used in zapi_test.go. It's in the test file so I think it's fine that it's exported for now (has always been that way).

PR #298, opened 2023-10-13
delete route API in sysrib

wenovus:


sysrib/server.go
```diff
@@ -126,6 +122,17 @@ func (d *dplane) ProgramRoute(r *ResolvedRoute) error {
 	return err
 }
 
+// deprogramRoute de-programs the route in the dataplane, returning an error on failure.
+func (d *dplane) deprogramRoute(r *ResolvedRoute) error {
```
DanG100, line 126: this should accept a context

wenovus: Addressed in next PR: https://github.com/openconfig/lemming/pull/301

PR #299, opened 2023-10-13
Add common pre-ingress, ingress, egress stages to pipeline and bind acls

DanG100:
* support for acl types
* new action table type
* add pre-ingress, ingress, egress stages to pipleine


proto/forwarding/forwarding_table.proto
```diff
@@ -175,6 +181,19 @@ message BridgeTableDesc {
...
+message ActionEntryDesc {
+  enum Type {
+    TYPE_UNSPECIFIED = 0;
+    TYPE_PREPEND = 1;
+    TYPE_APPEND = 2;
+  }
+  string id = 1;
+  Type insert_type = 2;
+}
+
```
wenovus, line 196: where is the data to be inserted? Should this be called InsertionEntryDesc? Action is the most generic term.

dataplane/standalone/saiserver/acl.go
```diff
@@ -35,6 +37,7 @@ type aclDataplaneAPI interface {
 	ID() string
 	TableCreate(context.Context, *fwdpb.TableCreateRequest) (*fwdpb.TableCreateReply, error)
 	TableEntryAdd(context.Context, *fwdpb.TableEntryAddRequest) (*fwdpb.TableEntryAddReply, error)
+	PortIDToNID(port string) (uint64, bool)
```
wenovus, line 40: What's NID?

wenovus (approved):

wenovus:
Also the lint error.

proto/forwarding/forwarding_table.proto
DanG100: its ActionEtntryDesc because the table is call ActionTable (its a list of actions). this follow pattern of other tables

dataplane/standalone/saiserver/acl.go
DanG100: numeric id

PR #303, opened 2023-10-16
Implement bulk neigh API

DanG100:
With the change, i was able to get the e2e test work.

Will add missing features + UT in next PR

### DIFF
--- a/dataplane/standalone/saiserver/routing.go
+++ b/dataplane/standalone/saiserver/routing.go
@@ -71,6 +71,19 @@ func (n *neighbor) CreateNeighborEntry(ctx context.Context, req *saipb.CreateNei
 	return &saipb.CreateNeighborEntryResponse{}, nil
 }
 
+// CreateNeighborEntries adds multiple neighbors to the neighbor table.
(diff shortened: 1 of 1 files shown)


dataplane/standalone/saiserver/routing.go
```diff
@@ -71,6 +71,19 @@ func (n *neighbor) CreateNeighborEntry(ctx context.Context, req *saipb.CreateNei
 	return &saipb.CreateNeighborEntryResponse{}, nil
 }
 
+// CreateNeighborEntries adds multiple neighbors to the neighbor table.
+func (n *neighbor) CreateNeighborEntries(ctx context.Context, re *saipb.CreateNeighborEntriesRequest) (*saipb.CreateNeighborEntriesResponse, error) {
+	resp := &saipb.CreateNeighborEntriesResponse{}
+	for _, req := range re.GetReqs() {
+		res, err := attrmgr.InvokeAndSave(ctx, n.mgr, n.CreateNeighborEntry, req)
```
wenovus, line 78: When is Interceptor not sufficient for storing attributes in the request?

wenovus (approved):

dataplane/standalone/saiserver/routing.go
DanG100: the interceptor is only for gRPC. when invoking server funcs directly need to use InvokeAndSave

PR #306, opened 2023-10-17
Add support for required ACL fields

DanG100:


dataplane/forwarding/protocol/packet/ip4_test.go
```diff
@@ -31,55 +31,46 @@ import (
 var ip4 = []byte{0x45, 0x01, 0x00, 0x16, 0x00, 0x00, 0x00, 0x00, 0x08, 0xff, 0x00, 0x00, 0x01, 0x02, 0x03, 0x04, 0x0a, 0x0b, 0x0c, 0x0d, 0x00, 0x00}
```
wenovus, line 31: Change to 0x31 to test masking of last two bits.

dataplane/standalone/saiserver/acl.go
```diff
@@ -169,6 +169,99 @@ func (a *acl) CreateAclEntry(ctx context.Context, req *saipb.CreateAclEntryReque
 			Masks:   binary.BigEndian.AppendUint64(nil, math.MaxUint64),
 		})
 	}
+	if req.GetFieldAclIpType() != nil {
+		fieldMask := &fwdpb.PacketFieldMaskedBytes{
```
wenovus, line 173: which field's mask? Scanning the rest of the code lends me to think this is creating masks depending on what is specified in the request, is that correct?  Perhaps explain how the masks and bytes are being applied to make it clearer what is happening here.

wenovus (approved):

dataplane/forwarding/protocol/packet/ip4_test.go
DanG100: reverted this, the IP_POS already exists and covers this field

PR #309, opened 2023-10-20
Support BGP Route Deletion

wenovus:
* RedistributeDel non-connected routes from RIB to ZAPI clients.
* Delete routes from RIB when ZAPI RouteDelete message is received.


DanG100 (approved):

wenovus:
The lint error is very strange because WildcardQuery has a PathStruct() method. Let's see if this continues.

PR #315, opened 2023-10-25
Move to Go 1.21 and update deps

wenovus:
Credit to @DanG100 for fixing the bazel import issues.

NOTE: I disabled the gNSI integration test. Not sure how to fix that given Ondatra has removed support for CustomData.


DanG100 (approved):

wenovus:
Cannot repro the route propagation test failure locally (IPv6 prefix is not propagated). Not sure if it's because we're not waiting long enough. If this recurs enough will take a look.

PR #317, opened 2023-10-27
Cleanup packet io and notifications

DanG100:
* Use streaming RPC for packet io
* Remove reference to address

This will be used in the future to support packet io operations


proto/forwarding/forwarding_service.proto
```diff
@@ -110,12 +107,12 @@ service Forwarding {
...
 
-// PacketSink is a service that processes packets on the specified port.
-service PacketSink {
-  // PacketInject injects a packet into the specified port.
-  rpc PacketInject(PacketInjectRequest) returns (PacketInjectReply) {}
+  // PacketInject injects of packets into any port in the context.
+  rpc PacketInject(stream PacketInjectRequest) returns (PacketInjectResponse) {}
+
+  // PacketEject receives all packets sent to the CPU port.
+  rpc PacketEject(PacketEjectRequest) returns (stream PacketEjectResponse) {}
```
wenovus, line 115: Can you describe this a little more? Why is inject "any" port but eject only the "cpu" port? Maybe PacketSinkRequest is a better name? Just a suggestion

dataplane/forwarding/fwdport/ports/cpu.go
```diff
@@ -121,21 +121,19 @@ func (p *cpuPort) punt(v interface{}) {
 			Bytes:   value,
 		})
 	}
-	request := &fwdpb.PacketInjectRequest{
-		ContextId:    &fwdpb.ContextId{Id: p.ctx.ID},
+	request := &fwdpb.PacketEjectResponse{
```
wenovus, line 124: request -> response

wenovus (approved):

proto/forwarding/forwarding_service.proto
DanG100: renamed to PacketSinkSubscribe, so it matches close the NotifySubscibe. as they operate similar

dataplane/forwarding/fwdport/ports/cpu.go
DanG100: done

PR #324, opened 2023-11-08
Quick fix for hostif index calculation

DanG100:
Plan for a longer term in this in works that doesn't make assumption of about interface naming

### DIFF
--- a/dataplane/standalone/sai/hostif.cc
+++ b/dataplane/standalone/sai/hostif.cc
@@ -70,7 +70,7 @@ lemming::dataplane::sai::CreateHostifRequest convert_create_hostif(
         int count = sscanf(attr_list[i].value.chardata, "Ethernet%d", &idx);
         if (count == 1) {
           std::ostringstream s;
-          s << "ip link set eth" << idx / 4 << " name "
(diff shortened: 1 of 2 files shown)


dataplane/standalone/sai/hostif.cc
```diff
@@ -68,11 +66,15 @@ lemming::dataplane::sai::CreateHostifRequest convert_create_hostif(
         msg.set_obj_id(attr_list[i].value.oid);
         break;
       case SAI_HOSTIF_ATTR_NAME: {
-        std::ostringstream s;
-        s << "ip link set eth" << nextIdx++ << " name "
-          << attr_list[i].value.chardata;
-        LOG(INFO) << s.str();
-        system(s.str().c_str());
+        int idx;
```
wenovus, line 69: If short-term then add a TODO on the long-term fix?

DanG100: https://github.com/openconfig/lemming/blob/main/dataplane/standalone/apigen/ccgen/ccgen.go#L161

wenovus (approved):

PR #325, opened 2023-11-08
CPU port implementation part one

DanG100:
Background for what will implemented in next PRs:
* The will be grpc client running in VM that is the "CPU port"
* The cpu port receives packets and port information from the dataplane
* The cpu port uses the packet egress ID and the port information to write the frames to the correct places


dataplane/forwarding/fwdport/ports/cpu_test.go
```diff
@@ -106,7 +106,7 @@ func TestCpuWrite(t *testing.T) {
 
 	// Verify that the packet was received and the parsed fields only have
 	// the ETHER_TYPE set to ARP.
-	list := ps.lastPacket.GetParsedFields()
+	list := ps.lastPacket.GetPacket().GetParsedFields()
```
wenovus, line 109: is `lastPacket` still an accurate name?

dataplane/forwarding/fwdport/ports/genetlink.go
```diff
@@ -0,0 +1,114 @@
...
+	if p.output, err = fwdaction.NewActions(genUpd.Genetlink.GetOutputs(), p.ctx); err != nil {
+		return fmt.Errorf("ports: output actions for port %v failed, err %v", p, err)
+	}
+	return nil
+}
+
+// Write writes a packet out. If successful, the port returns
+// fwdaction.CONSUME.
+func (p *genetlinkPort) Write(fwdpacket.Packet) (fwdaction.State, error) {
+	return fwdaction.CONSUME, nil
```
wenovus, line 83: Will the grpc server be writing out packets from here?

wenovus (approved):

dataplane/forwarding/fwdport/ports/cpu_test.go
DanG100: renamed

dataplane/forwarding/fwdport/ports/genetlink.go
DanG100: no, from the cpu port

PR #331, opened 2023-11-16
Part one of engine deprecation

DanG100:
Maintaining both engine and saiserver API is becoming difficult. 

First step of deprecation making the saiserver not rely on engine.
Next step will migrating the gnmi reconcilers to use the sai APIs

Review note: for now this is introducing a lot of code duplication, but this is short term problem


dataplane/internal/engine/engine.go
```diff
@@ -78,10 +77,11 @@ type Engine struct {
 	// internalToExternalID is a map from the internal port id to it's corresponding external port.
 	internalToExternalID map[string]string
 	cancelFn             func()
+	createTables         bool
 }
 
 // New creates a new engine and sets up the forwarding tables.
-func New(ctx context.Context) (*Engine, error) {
+func New(ctx context.Context, createTables bool) (*Engine, error) {
```
wenovus, line 84: Add comment for why you might not want to create tables.

dataplane/standalone/proto/common.proto
wenovus, line 1: Curious what's with the number changes?

wenovus (approved):

dataplane/internal/engine/engine.go
DanG100: added

dataplane/standalone/proto/common.proto
DanG100: the c++ to proto conversion code does enum conversion as raw int assignement so the numbers need to match up correctly

PR #343, opened 2024-01-02
Add comment to gnmi.go to suggest using gRPC interceptor when creating grpc.Server object.

wenovus:
Follow-up from https://github.com/openconfig/lemming/issues/263#issuecomment-1874642307

### DIFF
--- a/gnmi/gnmi.go
+++ b/gnmi/gnmi.go
@@ -77,7 +77,7 @@ type Server struct {
 //
 // To avoid having to specify the Target field in gNMI requests in order to comply
 // with https://www.openconfig.net/docs/gnmi/gnmi-specification/#2221-path-target,
-// register the gRPC interceptor using
+// register the gRPC interceptor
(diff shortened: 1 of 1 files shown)


gnmi/gnmi.go
```diff
@@ -75,6 +75,12 @@ type Server struct {
 
 // New creates and registers a reference gNMI server on the given gRPC server.
 //
+// To avoid having to specify the Target field in gNMI requests in order to comply
```
DanG100, line 78: would it be possible to change the implementation so this isn't needed?
before to calling cache Subscribe populate the target?

DanG100 (approved):

gnmi/gnmi.go
wenovus: Yes that might be possible and would increase the usability of the fake gNMI server. I'll refactor the code from the interceptor instead of this change in a different PR.

PR #344, opened 2024-01-10
Initial impl of a local ondatra binding

DanG100:
* Better config solution for dataplane
* Support inject arbitrary port implementation in the dataplane
* Add local binding that starts lemmings in the memory

In a follow up PR, i'll implement the dial methods and expand the test to something real


integration_tests/dataplane/basictraffic/basic_traffic_test.go
```diff
@@ -0,0 +1,18 @@
+package basictraffic
```
wenovus, line 1: This is a debugging test right?

internal/binding/local.go
```diff
@@ -0,0 +1,226 @@
...
+
+	opb "github.com/openconfig/ondatra/proto"
+
+	saipb "github.com/openconfig/lemming/dataplane/proto"
+	fwdpb "github.com/openconfig/lemming/proto/forwarding"
+)
+
+type LocalBind struct {
+	binding.Binding
+}
```
wenovus, line 44: comment

internal/binding/local.go
```diff
@@ -0,0 +1,226 @@
...
+
+type localLemming struct {
+	binding.AbstractDUT
+	l     *lemming.Device
+	dutID string
+}
+
+const (
+	gnmiPort      = 9339
+	gribiPort     = 9559
```
wenovus, line 65: Should this not be 9340? https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt

internal/binding/local.go
```diff
@@ -0,0 +1,226 @@
...
+	bgpPort       = 1179
+	dataplanePort = 9999
```
wenovus, line 67: I'm guessing pedantically this should be 49152-65535 for a dynamic, private or [ephemeral ports](https://en.wikipedia.org/wiki/Ephemeral_port)

internal/binding/local.go
```diff
@@ -0,0 +1,226 @@
...
+type localLemming struct {
+	binding.AbstractDUT
+	l     *lemming.Device
+	dutID string
+}
+
+const (
+	gnmiPort      = 9339
+	gribiPort     = 9559
+	bgpPort       = 1179
```
wenovus, line 66: For this local binding, are they all running in the same network namespace? If so should this be configurable since each lemming will spin up a BGP server?

integration_tests/dataplane/basictraffic/basic_traffic_test.go
DanG100: it tests that the reservation works, added todo to extend to real test

internal/binding/local.go
DanG100: they are running on different IPs

dataplane/dplaneopts/dplaneopts.go
```diff
@@ -0,0 +1,79 @@
...
+// limitations under the License.
+
+package dplaneopts
+
+import fwdpb "github.com/openconfig/lemming/proto/forwarding"
+
+// Options configures the dataplane
+type Options struct {
+	// Addr is the address of the gRPC server.
+	Addr string
```
wenovus, line 22: AddrPort

dataplane/dplaneopts/dplaneopts.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// Option exposes additional configuration for the dataplane.
+type Option func(*Options)
+
+// WithAddrPort sets the address of the dataplane gRPC server
+// Default: 127.0.0.1:0
+func WithAddrPort(addr string) Option {
+	return func(o *Options) {
+		o.Addr = addr
```
wenovus, line 38: addr -> addrPort

dataplane/forwarding/infra/fwdcontext/context.go
```diff
@@ -62,6 +74,8 @@ type Context struct {
 
 	eventMu     sync.Mutex // Mutex protecting the event notification
 	nextEventID uint64     // Id of the next event id
+
+	FakePortManager FakePortManager
```
wenovus, line 78: comment

wenovus (approved):

PR #353, opened 2024-01-31
Avoid redistributing BGP routes back to BGP speaker in a loop.

wenovus:
BGP routes distributed from GoBGP -> sysRIB must not be redistributed back into GoBGP. Otherwise GoBGP thinks it's a locally-originated route and will prefer it over all other routes, and any BGP path attributes are also erased in the process.

BGP-triggered GUE had to be tweaked because OTG tears down the BGP session on PushConfig, so every instance of the traffic test has to wait for the BGP session to re-establish first. This issue was masked by the above bug in sysRIB.

### DIFF
--- a/internal/debug/debug.go
+++ b/internal/debug/debug.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package debug contains flag for turning on or off debug messages.
+// Package debug contains flags for turning on or off debug messages.
 package debug
 
 const (
@@ -22,4 +22,6 @@ const (
 	// TAPPortPacketTrace turns on packet tracing for lemming's TAP interface ports
 	// that communicate between each external port and lemming's tasks.
 	TAPPortPacketTrace = true
+	// SysRIB controls whether debug messages for the main system RIB are printed.
+	SysRIB = false
 )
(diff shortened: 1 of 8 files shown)


wenovus:
@DanG100 The errors are due to
```
Build has expired and capabilities of product has been reduced, please contact Keysight Support
```
The tests (apart from aggregate) pass for me locally.

I prefer to have this merged soon because it is a bug in sysrib.

internal/debug/debug.go
```diff
@@ -22,4 +22,6 @@ const (
 	// TAPPortPacketTrace turns on packet tracing for lemming's TAP interface ports
 	// that communicate between each external port and lemming's tasks.
 	TAPPortPacketTrace = true
+	// SysRIB controls whether debug messages for the main system RIB are printed.
+	SysRIB = false
```
DanG100, line 26: maybe add TODO for use viper for this you don't have change src code

DanG100 (approved):

internal/debug/debug.go
wenovus: Do you mean so that the flags don't all have to be added to this file, and by changing topology.txtpb we get access to different systems' debug messages?

DanG100: yes

PR #365, opened 2024-02-27
Run tests in parallel

DanG100:


dataplane/dplanerc/interface.go
```diff
@@ -404,6 +404,17 @@ func (ni *Reconciler) addLAGMember(ctx context.Context, intf ocInterface, member
 		if err := ni.ifaceMgr.LinkSetDown(memberLink); err != nil {
 			log.Warningf("failed to set link %v down: %v", intf.name, err)
 		}
+		for i := 0; i < 3; i++ { // Ensure link is down before continuing.
```
guoshiuan, line 407: How is 3 selected? can we make it a constant?

dataplane/dplanerc/interface.go
```diff
@@ -404,6 +404,17 @@ func (ni *Reconciler) addLAGMember(ctx context.Context, intf ocInterface, member
...
+			memberLink, err := ni.ifaceMgr.LinkByIndex(memberData.hostifIfIndex)
+			if err != nil {
+				break
```
guoshiuan, line 410: I suppose the error is due to non-existing link. Should we return error or we just don't care?

dataplane/dplanerc/interface.go
DanG100: make a const

dataplane/dplanerc/interface.go
DanG100: changed to continue

dataplane/dplanerc/interface.go
```diff
@@ -404,10 +406,10 @@ func (ni *Reconciler) addLAGMember(ctx context.Context, intf ocInterface, member
 		if err := ni.ifaceMgr.LinkSetDown(memberLink); err != nil {
 			log.Warningf("failed to set link %v down: %v", intf.name, err)
 		}
-		for i := 0; i < 3; i++ { // Ensure link is down before continuing.
+		for i := 0; i < ifaceDownRetries; i++ { // Ensure link is down before continuing.
 			memberLink, err := ni.ifaceMgr.LinkByIndex(memberData.hostifIfIndex)
 			if err != nil {
-				break
+				continue
```
guoshiuan, line 412: add a sleep here too.

guoshiuan (approved):

PR #366, opened 2024-03-04
PacketIO proto API, metadata, and ports

DanG100:


dataplane/internal/kernel/genetlink.go
```diff
@@ -57,17 +58,23 @@ func NewGenetlinkPort(family, group string) (*GenetlinkPort, error) {
 	}, nil
 }
 
+type PacketMetadata struct {
+	SrcIfIndex int
+	DstIfIndex int
+	Context    int
```
guoshiuan, line 64: Just curious, what is Context? an int?

dataplane/internal/kernel/genetlink.go
```diff
@@ -57,17 +58,23 @@ func NewGenetlinkPort(family, group string) (*GenetlinkPort, error) {
...
+}
+
 // Writes writes a layer2 frame to the port.
-func (p GenetlinkPort) Write(frame []byte) error {
+func (p GenetlinkPort) Write(frame []byte, md *PacketMetadata) (int, error) {
 	// TODO: Support src, dst, and context
```
guoshiuan, line 69: Remove this comment.

proto/forwarding/forwarding_port.proto
```diff
@@ -222,3 +223,35 @@ message PortStateRequest {
...
+    ContextId context_id = 1;
+    google.rpc.Status status = 2;
+  }
+}
+
+message HostPortControlMessage {
+  PortDesc port = 1;
+  uint64 port_id = 2;
+  PortId dataplane_port = 3; // ID of the dataplane port related to this one.
+  bool create = 4;
```
guoshiuan, line 238: How is create used?

dataplane/internal/kernel/genetlink.go
DanG100: just an extra value that the SAI client can configure

proto/forwarding/forwarding_port.proto
DanG100: Set to true if an hostif should be created, false for deletion.

guoshiuan (approved):

PR #379, opened 2024-03-18
Correctly populate community route attributes in BGP RIBs.

wenovus:
Sync fix from upstream GoBGP and add missing unit tests.

Also added sanity community set checks into integration_test/prefix_test to avoid creating another integration test that tests the same logic. Added TODO for race conditions discovered that are triggered by subtests when BGP policy and routes are cleared and re-created. This will block policy logic tests that involve re-advertising the same route under different policies, but will work for tests that are checking a static set of policies, which is the current set of tests I'm looking at to support.


bgp/gobgp.go
```diff
@@ -457,6 +481,18 @@ func (t *bgpTask) updateRIBs(ctx context.Context) error {
 						//
 						// Note that path.NeighborIp is <nil> for some reason so have to use neigh.
 						t.ribattrs.populateRIBAttrs(path, bgpRIB, neighContainer.GetOrCreateAdjRibOutPre().GetOrCreateRoute(route.Prefix, uint32(i)))
+					}
+				}
+			})
+			t.queryTable(ctx, neigh, "adj-rib-out-post-v6", true, api.TableType_ADJ_OUT, api.Family_AFI_IP6, func(routes []*api.Destination) {
```
DanG100, line 487: no action required: would be possible to construct the string from the arguments to this func? it seems like the name is TableType + Family + postPolicy bool

DanG100 (approved):

bgp/gobgp.go
wenovus: Done.

PR #386, opened 2024-03-22
Integration test to tunnel + remove tunnel and remove tunnel term

DanG100:


dataplane/saiserver/attrmgr/attrmgr.go
```diff
@@ -217,7 +217,6 @@ func (mgr *AttrMgr) PopulateAttributes(req, resp proto.Message) error {
 }
 
 // PopulateAllAttributes fills the resp with the requests attributes.
```
guoshiuan, line 219: Please update the comment too.

guoshiuan (approved):

dataplane/saiserver/attrmgr/attrmgr.go
DanG100: done

PR #387, opened 2024-03-22
Fix local bind and support multiple readers and writers to a port

DanG100:


internal/binding/local.go
```diff
@@ -412,127 +417,196 @@ func (lb *LocalBind) createATE(_ context.Context, ate *opb.Device, portMgr *ateP
...
+	dutID string
 }
 
-// dutPortMgr implements lemming's port creator interface.
-// Note: this interface is different from magna, to avoid making the two project depending on eachother.
-type dutPortMgr struct {
-	ports map[string]*chanPort
-	// TODO: lastCreatedPort is a hack because we don't currently model hardware ports correctly in lemming.
-	lastCreatedPort *chanPort
+// CreatePort impplements lemming's API for creating port.
```
guoshiuan, line 528: implements

internal/binding/local.go
```diff
@@ -412,127 +417,196 @@ func (lb *LocalBind) createATE(_ context.Context, ate *opb.Device, portMgr *ateP
...
+			p.mu.RUnlock()
+		}
+	}()
 }
 
-// atePortMgr implement magna's port creator interface.
-// Note: this interface is different from lemming, to avoid making the two project depending on eachother.
-type atePortMgr struct {
-	ports map[string]*chanPort
+type portHandle struct {
```
guoshiuan, line 580: Can we have comment on this struct? look to me this is only for packet capture?

guoshiuan (approved):

internal/binding/local.go
DanG100: done

internal/binding/local.go
DanG100: done

PR #391, opened 2024-03-28
More tests with debugging.

guoshiuan:


integration_tests/dataplane/mymac/mymac_test.go
```diff
@@ -189,65 +186,147 @@ func clearMyMac(t *testing.T, dut *ondatra.DUTDevice) error {
...
+	macAddr     string
+	macAddrMask []byte
+}
+
+type myMacTest struct {
+	desc       string
+	clearMyMac bool
+	dstMAC     string
+	entries    []entry
+	passed     bool
```
DanG100, line 255: nit: rename passed to wantTraffic (or something like that wantNoDrop etc) since all tests should pass

integration_tests/dataplane/mymac/mymac_test.go
```diff
@@ -189,65 +186,147 @@ func clearMyMac(t *testing.T, dut *ondatra.DUTDevice) error {
...
+	priority    uint32
+	macAddr     string
+	macAddrMask []byte
+}
+
+type myMacTest struct {
+	desc       string
+	clearMyMac bool
+	dstMAC     string
+	entries    []entry
```
DanG100, line 254: why not just use the proto type CreateMyMac directly in the case

integration_tests/dataplane/mymac/mymac_test.go
```diff
@@ -189,65 +186,147 @@ func clearMyMac(t *testing.T, dut *ondatra.DUTDevice) error {
...
+			macAddr:     dutPort2.MAC,
+			macAddrMask: []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff},
+		}, {
+			priority:    2000,
+			macAddr:     specialMAC,
+			macAddrMask: []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff},
+		}},
+		passed: false,
+	}, {
+		desc:       "Traffic passed with special MAC address",
```
DanG100, line 295: nit: special is not descriptive, something like  "passed where dst mac not interface mac"

integration_tests/dataplane/mymac/mymac_test.go
guoshiuan: done

integration_tests/dataplane/mymac/mymac_test.go
guoshiuan: Done

integration_tests/dataplane/mymac/mymac_test.go
guoshiuan: Done

DanG100 (approved):

PR #392, opened 2024-03-29
Handle pktiohandler restarts and ports info to file

DanG100:


dataplane/saiserver/hostif.go
```diff
@@ -57,7 +59,19 @@ type hostif struct {
...
+	remoteHostifs    map[uint64]*pktiopb.HostPortControlMessage
+	remoteClosers    []func()
+	remotePortReq    func(msg *pktiopb.HostPortControlMessage) error
+}
+
+func (hostif *hostif) Reset() {
+	hostif.trapIDToHostifID = map[uint64]uint64{}
+	hostif.groupIDToQueue = map[uint64]uint32{}
+	hostif.remoteHostifs = map[uint64]*pktiopb.HostPortControlMessage{}
+	for _, closeFn := range hostif.remoteClosers {
```
guoshiuan, line 72: Do we need to clear remoteClosers too, and then re-init?

dataplane/saiserver/hostif.go
```diff
@@ -235,6 +249,8 @@ func (hostif *hostif) CreateHostif(ctx context.Context, req *saipb.CreateHostifR
 
 func (hostif *hostif) createRemoteHostif(ctx context.Context, req *saipb.CreateHostifRequest) (*saipb.CreateHostifResponse, error) {
 	id := hostif.mgr.NextID()
+	hostif.remoteMu.Lock()
```
guoshiuan, line 252: I think remoteMu is used to guard remoteHostifs[], should we move it closer to where remoteHostifs is acutally used?

guoshiuan (approved):

dataplane/saiserver/hostif.go
DanG100: done

dataplane/saiserver/hostif.go
DanG100: done

PR #395, opened 2024-04-04
Hash API

DanG100:


dataplane/saiserver/routing_test.go
```diff
@@ -226,15 +222,22 @@ func TestCreateNextHopGroupMember(t *testing.T) {
...
+			NextHopGroupId: proto.Uint64(2),
+			NextHopId:      proto.Uint64(3),
 			Weight:         proto.Uint32(3),
 		},
 	}}
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
 			dplane := &fakeSwitchDataplane{}
 			c, mgr, stopFn := newTestNextHopGroup(t, dplane)
+
```
guoshiuan, line 234: Can we add some comments on these numbers: 10 and 3?

guoshiuan (approved):

dataplane/saiserver/routing_test.go
DanG100: they are just random numbers i picked

PR #401, opened 2024-04-16
Support Minimum gNOI Reboot and CancelReboot Functionality

wenovus:
* Populate minimum paths in CHASSIS component type to have `gNOI-3.1: Complete Chassis Reboot` pass.
* For Reboot and CancelReboot, only the delay parameter is supported.
* Re-enabled fakedevice.NewCurrentTimeTask()

The Reboot/CancelReboot behaviour is taken from the comments in https://github.com/openconfig/gnoi/blob/main/system/system.proto. The implementation uses non-straightforward concurrency for cleaning-up concurrency signals, but I didn't see another way to do this for now since the spawned goroutine has a limited lifetime.

### DIFF
--- a/gnoi/gnoi.go
+++ b/gnoi/gnoi.go
@@ -16,8 +16,11 @@ package gnoi
 
 import (
 	"context"
+	"sync"
 	"time"
 
+	log "github.com/golang/glog"
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
 	bpb "github.com/openconfig/gnoi/bgp"
 	cmpb "github.com/openconfig/gnoi/cert"
 	diagpb "github.com/openconfig/gnoi/diag"
@@ -30,7 +33,11 @@ import (
 	otpb "github.com/openconfig/gnoi/otdr"
 	spb "github.com/openconfig/gnoi/system"
 	wrpb "github.com/openconfig/gnoi/wavelength_router"
+	"github.com/openconfig/lemming/gnmi/fakedevice"
+	"github.com/openconfig/ygnmi/ygnmi"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type bgp struct {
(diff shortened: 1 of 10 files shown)


gnoi/gnoi.go
```diff
@@ -75,12 +82,95 @@ type otdr struct {
...
 }
 
+func (s *system) Reboot(ctx context.Context, r *spb.RebootRequest) (*spb.RebootResponse, error) {
+	if r.Method == spb.RebootMethod_POWERUP {
+		return &spb.RebootResponse{}, nil
+	}
+
+	s.rebootMu.Lock()
+	defer s.rebootMu.Unlock()
+	if s.hasPendingReboot {
```
DanG100, line 113: i think this approach is fine: however a few alternatives
1. use an atomic.Bool for hasPendingReboot simplify/eliminate the use of mutexes for these RPCs?
2. instead of having multiple chans, have a func pointer called cancelReboot, assigned in Reboot and called in CancelReboot
3. initialize the chans in the Reboot and close instead of sending empty structs

gnoi/gnoi_test.go
```diff
@@ -0,0 +1,197 @@
...
+			t.Fatal(err)
+		}
+	})
+
+	closeProximityTests := []struct {
+		desc  string
+		delay uint64
+	}{{
+		desc:  "cancel-while-possibly-pending-1ns",
+		delay: 1e0,
```
DanG100, line 148: nit: use time.Duration in the test case and delay.Unix().Nanos() in the req

integration_tests/onedut_tests/complete_chassis_reboot/complete_chassis_reboot_test.go
```diff
@@ -0,0 +1,241 @@
...
+					}
+					prevTime = latestTime
+				}
+			}
+
+			startReboot := time.Now()
+			t.Logf("Wait for DUT to boot up by polling the telemetry output.")
+			for {
+				var currentTime string
+				t.Logf("Time elapsed %.2f seconds since reboot started.", time.Since(startReboot).Seconds())
```
DanG100, line 166: does this test make sense given we aren't actually rebooting the node

gnoi/gnoi.go
wenovus: Thanks for the suggestions. I tried each of these but didn't see a way to make them work for the reasons below. I've added comments in the code to justify using the data types chosen.

1. atomic.Bool doesn't quite work because the mutex also makes sure the sequence `check no pending reboots -> immediate reboot` is carried out atomically. If `.Load()` is called at this line, then it's possible that before the immediate reboot, that a different delayed reboot operation was scheduled such that `hasPendingReboot` is true when the immediate reboot is executed, essentially a TOCTTOU bug. Now this doesn't actually cause a practical problem in this case but semantically I think it's wrong so I prefer using the mutex.
2. I think this is an interesting idea. However the channels are needed to make CancelReboot block until the cancellation succeeds, otherwise if a reboot operation is issued too quickly after a cancel, then it might cause a conflict with the previously scheduled reboot that hasn't been fully cancelled yet.
3. I considered this, but because the spawned goroutine has a limited lifetime, I didn't see a clean way to re-create the channel for re-use after closing them. A channel cannot be closed twice.

gnoi/gnoi_test.go
wenovus: Done.

integration_tests/onedut_tests/complete_chassis_reboot/complete_chassis_reboot_test.go
wenovus: The test doesn't enforce any downtime so at least it passes :) -- this is more of a demonstration PR to show that lemming can implement portions of gNOI. If the need arises we'll have to simulate downtime.

DanG100 (approved):

PR #423, opened 2024-06-03
Make BGP prefix checking less flaky

wenovus:
Currently it's failing because the `Await` check before the `Get` wait for **any** prefix rather than the prefix `Get` is looking for. This makes some of the checks fail intermittently when multiple prefixes are checked at the same time.


DanG100:
this test is still failing on the presubmit

wenovus:
> this test is still failing on the presubmit

Yeah I'm not sure what's going on, the test consistently passes locally. Will dig through the logs.

DanG100 (approved):

PR #426, opened 2024-06-06
Update pathz.go

marcushines:


DanG100 (approved):

gnsi/pathz/pathz.go
```diff
@@ -54,8 +55,15 @@ func (s *Server) Rotate(rs pathzpb.Pathz_RotateServer) error {
 	s.rotationInProgress.Store(true)
 	defer s.rotationInProgress.Store(false)
 
+	ctx := rs.Context()
 	receivedUploadReq := false
 	for {
+		select {
```
rnhaddad, line 61: This will change the "behavior" in that, it is now up to the discretion of the client to close the stream. Is this intended? People had different opinions on the "Finalize" behavior, where, if the server has finalized, the stream had to be forcibly terminated by the server (which was inline with the return nil from the gRPC handler below).

If this is intentional, i.e. keeping the stream alive based on client intent, then that is ok.

PR #429, opened 2024-06-11
Refactor policytest to have one test case per prefix rather than (prefix, path) tuple

wenovus:
Summary: put `RouteTest`, `AlternateRouteTest`, `LongerPathRouteTest` into a single test case struct sharing the same prefix.

Previously each was treated as its own test, but since some tests are testing the interaction between different BGP advertisement paths for the same prefix, it makes more sense to group these three advertisement paths into the same struct sharing the prefix.

The outcome is that it makes it easier to only run a single test when debugging.


wenovus:
I'm trying to fix the lint error at https://github.com/openconfig/lemming/pull/430

wenovus:
@DanG100 ping on this one, thanks. It's a mechanical change, just refactoring some fields from one struct to another.

DanG100 (approved):

PR #434, opened 2024-06-13
Support BGP policy's origin attributes

wenovus:


wenovus:
BGP tests passing locally.

PR #439, opened 2024-06-24
ACL code gen enhancements

DanG100:


dataplane/apigen/ccgen/ccgen.go
```diff
@@ -290,6 +298,44 @@ func createCCData(meta *saiast.FuncMetadata, apiName string, sai *saiast.SAIAPI,
 		convertFn = nil
 	}
 
+	// Patches for non-standard APIS
+	if meta.TypeName == "ACL_TABLE" {
+		switch meta.Operation {
+		case createOp:
+			convertFn.AttrConvertInsert = `
+if (attr_list[i].id >= SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MIN && attr_list[i].id < SAI_ACL_TABLE_ATTR_USER_DEFINED_FIELD_GROUP_MAX) {
```
guoshiuan, line 306: Do we need to care about the else situation?

dataplane/apigen/ccgen/ccgen.go
```diff
@@ -416,46 +463,64 @@ var typeToUnionAccessor = map[string]*unionAccessor{
...
 		convertFromFunc: "convert_from_acl_field_data_ip_type",
+		convertToFunc:   "convert_to_acl_field_data_ip_type",
+		protoAccessor:   "ip_type",
 		aType:           acl,
 	},
 	"sai_acl_field_data_t sai_uint8_t": {
 		accessor:        "u8",
 		convertFromFunc: "convert_from_acl_field_data",
+		convertToFunc:   "convert_to_acl_field_data_u8",
+		protoAccessor:   "uint",
```
guoshiuan, line 495: Both u8 and u16 are both handled by uint proto accessor? Do we need to make it more specific?

dataplane/apigen/ccgen/ccgen.go
```diff
@@ -416,46 +463,64 @@ var typeToUnionAccessor = map[string]*unionAccessor{
...
 		convertFromFunc: "convert_from_acl_field_data",
+		convertToFunc:   "convert_to_acl_field_data_u16",
+		protoAccessor:   "uint",
 		aType:           acl,
 	},
 	"sai_acl_field_data_t sai_ip6_t": {
 		accessor:        "ip6",
 		convertFromFunc: "convert_from_acl_field_data_ip6",
+		convertToFunc:   "convert_to_acl_field_data_ip6",
+		protoAccessor:   "ip",
```
guoshiuan, line 509: Same as above, looks like "ip" handles both ip4 and ip6.

dataplane/apigen/protogen/protogen.go
```diff
@@ -494,6 +494,7 @@ type saiTypeInfo struct {
 	Repeated   bool
 	ProtoType  string
 	MessageDef string
+	Required   bool
```
guoshiuan, line 497: Does it mean a required protobuf field?

dataplane/standalone/sai/common.cc
```diff
@@ -214,8 +214,94 @@ lemming::dataplane::sai::AclFieldData convert_from_acl_field_data_ip_type(
...
   lemming::dataplane::sai::AclFieldData out;
   out.set_enable(in.enable);
-  out.set_data_ip_type(
-      static_cast<lemming::dataplane::sai::AclIpType>(type + 1));
+  out.set_data_ip_type(convert_sai_acl_ip_type_t_to_proto(type));
   out.set_mask_int(mask);
   return out;
 }
+
+////
```
guoshiuan, line 222: Remove

guoshiuan (approved):

dataplane/apigen/ccgen/ccgen.go
DanG100: no, only need the special logic for these cases

dataplane/apigen/ccgen/ccgen.go
DanG100: protobuf only has 32 and 54 bit ints, so u16 and and u32 use the proto field

dataplane/apigen/ccgen/ccgen.go
DanG100: the protobuf ip field is bytes can be either v4 or v6, sai uses different types for v4 and v6

dataplane/apigen/protogen/protogen.go
DanG100: it means it shouldn't have the `optional` keyword

dataplane/standalone/sai/common.cc
DanG100: done

PR #440, opened 2024-06-25
UDF ACL impl

DanG100:


dataplane/saiserver/acl.go
```diff
@@ -250,6 +250,42 @@ func (a *acl) createAclEntryFields(req *saipb.CreateAclEntryRequest, id uint64,
...
+		udfGroup := &saipb.UdfGroupAttribute{}
+		if err := a.mgr.PopulateAllAttributes(fmt.Sprint(table.UserDefinedFieldGroupMin[id]), udfGroup); err != nil {
+			return nil, err
+		}
+		for _, udfID := range udfGroup.GetUdfList() {
+			udf := &saipb.UdfAttribute{}
+			if err := a.mgr.PopulateAllAttributes(fmt.Sprint(udfID), udf); err != nil {
+				return nil, err
+			}
+			hg := fwdpb.PacketHeaderGroup_PACKET_HEADER_GROUP_L2
```
guoshiuan, line 267: With this L2 is the default value. Is this safe?

guoshiuan (approved):

dataplane/saiserver/acl.go
DanG100: yes, this is intentional

PR #447, opened 2024-07-10
Quick findtrace script

DanG100:


cmd/findtrace/findtrace.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2023 Google LLC
```
guoshiuan, line 1: 2024

cmd/findtrace/findtrace.go
```diff
@@ -0,0 +1,51 @@
...
+func main() {
+	scanner := bufio.NewScanner(os.Stdin)
+	var trace strings.Builder
+
+	for scanner.Scan() {
+		text := scanner.Text()
+		if strings.Contains(text, "Packet Trace:") {
+			str := trace.String()
+			if strings.Contains(str, os.Args[1]) {
+				out := strings.ReplaceAll(str, os.Args[1], color.RedString(os.Args[1]))
```
guoshiuan, line 35: What is the first argument? It will be great to have an example usage command when arg is not specified.

guoshiuan (approved):

cmd/findtrace/findtrace.go
DanG100: added usage comment

PR #450, opened 2024-07-12
Add support for plumbing gRIBI server options through lemming.

robshakir:
```
 * (M) lemming.go
 * (M) gribi/gribi.go
  - for some behaviours, there is a need to change the underlying RIB functionality
    in the gRIBI server as a lemming is created. this change plumbs some options
    through to allow lemmings to be created using this method.
```


robshakir:
###

PR #458, opened 2024-07-22
Merge all misc commands into single CLI

DanG100:


dataplane/luciusctl/luciusctl.go
```diff
@@ -12,21 +12,20 @@
...
 	"github.com/openconfig/lemming/dataplane/luciusctl/info"
 	"github.com/openconfig/lemming/dataplane/luciusctl/sai"
 )
 
-func main() {
-	cmd := cobra.Command{
-		Use: "luciusctl",
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "lucius",
```
alexmasi, line 27: should this still be luciusctl?

alexmasi (approved):
any backwards compatibility concerns?

dataplane/luciusctl/luciusctl.go
DanG100: i thought that was too many characters to type

DanG100:
> any backwards compatibility concerns?

i don't think any one other than me is using these tools to debug

PR #465, opened 2024-08-12
Support log action and filter IPs

DanG100:


dataplane/saiserver/ports.go
```diff
@@ -92,6 +92,7 @@ func getPreIngressPipeline() []*fwdpb.ActionDesc {
 func getL3Pipeline() []*fwdpb.ActionDesc {
 	return []*fwdpb.ActionDesc{
 		fwdconfig.Action(fwdconfig.LookupAction(IngressActionTable)).Build(),                                                                                    // Run ingress action.
+		fwdconfig.Action(fwdconfig.LookupAction(invalidIPTable)).Build(),                                                                                        // Do not forward packets with invalid addresses
```
guoshiuan, line 95: Should this check happen before IngressActionTable?

guoshiuan (approved):

dataplane/saiserver/ports.go
DanG100: i think the desired behavior is that an ingress ACL rule could punt the packet before it's dropped, so this needs to happen after

PR #470, opened 2024-09-03
Add protocol registry to the server.

guoshiuan:


dataplane/protocol/registry.go
```diff
@@ -62,6 +63,11 @@ func NewRegistry(psc pktiopb.PacketIO_CPUPacketStreamClient) (*Registry, error)
 	return pr, nil
 }
 
+// Context returns the context of psc.
```
DanG100, line 66: nit: instead of psc say the grpc client.

DanG100:
to fix the check. delete the build file in dataplane/protocol/ and run `make buildfile` it doesn't handle package renames gracefully, need to help it a bit

DanG100 (approved):

dataplane/protocol/registry.go
guoshiuan: done.

PR #475, opened 2024-09-12
Initial tracing support for lucius

DanG100:


dataplane/standalone/lucius/lucius.go
```diff
@@ -92,3 +122,78 @@ func start(port int) {
...
+	otel.SetTracerProvider(tracerProvider)
+
+	shutdownFuncs = append(shutdownFuncs, meterProvider.Shutdown)
+	otel.SetMeterProvider(meterProvider)
+
+	// Set up logger provider.
+	loggerProvider, err := newLoggerProvider(res)
+	if err != nil {
+		handleErr(err)
+		return
```
guoshiuan, line 173: Can we avoid naked return?

guoshiuan (approved):

PR #488, opened 2024-10-24
Support copy action in ACL

DanG100:


dataplane/saiserver/acl.go
```diff
@@ -120,6 +121,12 @@ func (a *acl) CreateAclTableGroupMember(_ context.Context, req *saipb.CreateAclT
 // CreateAclTable is noop as the table is already created in the group.
 func (a *acl) CreateAclTable(context.Context, *saipb.CreateAclTableRequest) (*saipb.CreateAclTableResponse, error) {
 	id := a.mgr.NextID()
+
+	a.mgr.StoreAttributes(id, &saipb.AclTableAttribute{
+		AvailableAclEntry:   proto.Uint32(10000),
```
guoshiuan, line 126: These numbers are also used in switch.go and switch_test.go. Should we make these numbers const?

guoshiuan (approved):

dataplane/saiserver/acl.go
DanG100: they could be different, i just picked the same numbers since they are meaningless, there is no actual limit to acl entries

PR #493, opened 2024-11-01
Support route metadata

DanG100:


dataplane/saiserver/routing.go
```diff
@@ -510,6 +510,8 @@ func newRoute(mgr *attrmgr.AttrMgr, dataplane switchDataplaneAPI, s *grpc.Server
 	return r
 }
 
+const routeDstMeta = 0
```
guoshiuan, line 513: So we always have/care only one instance?

guoshiuan (approved):

dataplane/saiserver/routing.go
DanG100: the PACKET_ATTRIBUTE_32 is special, its actually a map, i'm using is this const as the key to for that map. added comment

dataplane/saiserver/acl_test.go
```diff
@@ -125,6 +125,10 @@ func TestCreateAclEntry(t *testing.T) {
 				Data: &saipb.AclFieldData_DataUint{DataUint: 0x01},
 				Mask: &saipb.AclFieldData_MaskUint{MaskUint: 0xff},
 			},
+			FieldRouteDstUserMeta: &saipb.AclFieldData{
+				Data: &saipb.AclFieldData_DataUint{DataUint: 1},
```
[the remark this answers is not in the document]
DanG100: done

PR #507, opened 2024-12-12
Test runner + example for dataplane tests

DanG100:


integration_tests/saiutil/saiutil.go
```diff
@@ -103,3 +131,138 @@ func CreateNeighbor(t testing.TB, dut *ondatra.DUTDevice, ip string, dmac string
 		t.Fatal(err)
 	}
 }
+
+type InterfaceRef struct {
```
guoshiuan, line 135: Any reason not to use NetworkInstance_Afts_NextHop_InterfaceRef?

integration_tests/saiutil/saiutil.go
```diff
@@ -30,6 +45,19 @@ func dataplaneConn(t testing.TB, dut *ondatra.DUTDevice) *grpc.ClientConn {
 	return conn
 }
 
+type configer struct {
+	cfg   func(t testing.TB, s *Suite, dut *ondatra.DUTDevice)
```
guoshiuan, line 49: Should these funcs return error?

integration_tests/saiutil/saiutil.go
```diff
@@ -103,3 +131,138 @@ func CreateNeighbor(t testing.TB, dut *ondatra.DUTDevice, ip string, dmac string
...
+		for pre, entry := range aft.Ipv6Entry {
+			CreateRoute(t, dut, pre, s.oc2SAINextHopGrp[entry.GetNextHopGroup()], s.oc2SAIVRF[vrf])
+		}
+	}
+	c.uncfg = func(t testing.TB, s *Suite, dut *ondatra.DUTDevice) {}
+
+	return c
+}
+
+func ConfigRIF(portID string, mac string, vlan string) *configer {
```
guoshiuan, line 193: vlan is not used, is it a place holder?

guoshiuan, line 193: Should vlan be VRF (see the reference in subintf_test.go)?

integration_tests/dataplane/subintf/subintf_test.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+				gopacket.Payload{},
+			},
+		},
+		Out: &saiutil.Packet{
+			Port: "port2",
+			Layers: []gopacket.SerializableLayer{
+				&layers.Ethernet{
+					SrcMAC:       saiutil.MustParseMac(t, "10:10:10:10:10:11"),
+					DstMAC:       saiutil.MustParseMac(t, "10:10:10:10:10:11"),
```
guoshiuan, line 113: why the dst MAC is the same with src MAC?

integration_tests/saiutil/saiutil.go
DanG100: there are other InterfaceRef types in the OC, this way we always use the same type

integration_tests/saiutil/saiutil.go
DanG100: these are test helpers so the fail the test instead of returning err

integration_tests/saiutil/saiutil.go
DanG100: had some local changes i forgot to commit. this should be VRF

DanG100: thanks fixed

integration_tests/dataplane/subintf/subintf_test.go
DanG100: changed for clarity

guoshiuan (approved):

PR #520, opened 2025-01-07
Add support for link up/down for packet handlers

DanG100:


dataplane/kernel/genetlink/genetlink.go
```diff
@@ -107,6 +107,10 @@ func (p GenetlinkPort) Delete() error {
 	return nil
 }
 
+func (t GenetlinkPort) SetAdminState(up bool) error {
+	return nil
```
guoshiuan, line 111: Where do we set the state?

DanG100: this interface type doesn't need it

PR #525, opened 2025-01-08
Send request to set oper status state for host interfaces.

royyi8:


dataplane/saiserver/hostif.go
```diff
@@ -91,6 +91,17 @@ func (hostif *hostif) CreateHostif(ctx context.Context, req *saipb.CreateHostifR
 		PortId:        id,
```
DanG100, line 91: can you add `Op: pktiopb.PortOperation_PORT_OPERATION_CREATE` to this struct. eventually we'll remove the Create field

dataplane/saiserver/hostif.go
```diff
@@ -91,6 +91,17 @@ func (hostif *hostif) CreateHostif(ctx context.Context, req *saipb.CreateHostifR
...
 	}
 
+	oper_status := pktiopb.PortOperation_PORT_OPERATION_SET_DOWN
```
DanG100, line 94: go is usually camel cased, rename `operStatus`

DanG100 (approved):

PR #526, opened 2025-01-14
Make cpu stream more resilient

DanG100:


dataplane/saiserver/hostif_test.go
```diff
@@ -209,6 +218,112 @@ func TestRemoveHostif(t *testing.T) {
...
+				InputPort: 2,
+			},
+		}
+		if d := cmp.Diff(got, want, protocmp.Transform(), protocmp.IgnoreFields(&pktiopb.Packet{}, protoreflect.Name("frame"))); d != "" {
+			t.Errorf("PacketStream() failed: diff(-got,+want)\n:%s", d)
+		}
+	})
+}
+
+func testPacket(t testing.TB, nid uint64) fwdpacket.Packet {
```
guoshiuan, line 285: What's the purpose of this test? Look to me it only tests whether packet field update is successful?

guoshiuan (approved):

dataplane/saiserver/hostif_test.go
DanG100: this is just a helper to create a packet, renamed func

PR #534, opened 2025-01-29
Allow passing extra to all lemmings

DanG100:


operator/main.go
```diff
@@ -64,6 +66,7 @@ func main() {
 		"If set the metrics endpoint is served securely")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
+	flag.StringVar(&extraArgs, "extra-args", "", "Additional args to pass to every instance of lemming")
```
guoshiuan, line 69: Looks to me it is a comma-sep string, but I'd like to know the detailed format. For example, is it a comma-sep string of "key=value", "--key=value", or it only supports bool style var?

guoshiuan (approved):

operator/main.go
DanG100: add more details

DanG100: added more details

PR #542, opened 2025-02-12
Clean up instrumentation

DanG100:


dataplane/saiserver/saiserver.go
```diff
@@ -233,5 +235,12 @@ func New(ctx context.Context, mgr *attrmgr.AttrMgr, s *grpc.Server, opts *dplane
 	saipb.RegisterSystemPortServer(s, srv.systemPort)
 	saipb.RegisterTamServer(s, srv.tam)
 
+	m := otel.GetMeterProvider().Meter("openconfig/lemming/dataplane/saiserver")
+	c, err := m.Int64Counter("lucius-instance")
+	if err != nil {
+		return nil, err
+	}
+	c.Add(ctx, 1)
```
guoshiuan, line 243: I suppose this increases the counter by 1, but why do we do that?

internal/cloudlog/cloudlog.go
```diff
@@ -19,60 +19,39 @@ import (
...
 	"go.opentelemetry.io/otel/trace"
 )
 
 // SetGlobal sets the global slog to use output a GCP cloud logging and stderr
-func SetGlobalLogger(ctx context.Context, project, logName string) error {
+func SetGlobalLogger(ctx context.Context, project, logName string, l slog.Level) func(context.Context) error {
 	cl, err := logging.NewClient(ctx, fmt.Sprintf("projects/%s", project))
 	if err != nil {
-		return err
+		return func(ctx context.Context) error { return nil }
```
guoshiuan, line 33: Now we are returning a func that returns no error, Is this WAI?

guoshiuan (approved):

dataplane/saiserver/saiserver.go
DanG100: we can aggregate all these counters to watch usage

internal/cloudlog/cloudlog.go
DanG100: i want this to be non-fatal if it fails, so it becomes a noop on error

PR #546, opened 2025-02-18
Fix target handling

DanG100:
also add more labels


internal/telemetry/telemetry.go
```diff
@@ -126,6 +127,7 @@ func Setup(ctx context.Context, options ...Option) (func(context.Context) error,
 	if err != nil {
 		return nil, err
 	}
+	fmt.Println(res)
```
guoshiuan, line 130: Remove this line if it is for debugging.

guoshiuan (approved):

PR #553, opened 2025-02-27
Improvements to fault server

DanG100:


fault/fault.go
```diff
@@ -48,28 +49,30 @@ type Interceptor struct {
...
-	msgID := uuid.New().String()
-
-	mpb, ok := msg.(proto.Message)
-	if !ok { // Do not intercept RPC where the response is not a protobuf message.
-		return msg, nil
-	}
-	mAny, err := anypb.New(mpb)
-	if err != nil {
-		return msg, oErr
+func (i *Interceptor) sendRecvFault(ch chan *faultMessage, rpcID string, msg any, msgType faultpb.MessageType, oErr error) (any, error) {
```
guoshiuan, line 61: Can we put comments on these parameters? Especially oErr and the return values?

fault/fault.go
```diff
@@ -139,47 +139,62 @@ type streamInt struct {
...
 	si := &streamInt{
 		ServerStream: stream,
 		int:          i,
 		fs:           sub,
 		rpcID:        uuid.New().String(),
 	}
-
-	return handler(srv, si)
+	hErr := handler(srv, si)
+	_, err := si.int.sendRecvFault(si.fs.originMsgCh, si.rpcID, nil, faultpb.MessageType_MESSAGE_TYPE_STREAM_END, hErr)
```
guoshiuan, line 177: Are we terminating the stream no matter what is returned from the handler?

guoshiuan (approved):

fault/fault.go
DanG100: no, when the handler exits, we allow a final opportunity to inject a fault.

PR #559, opened 2025-03-18
add reconciliation for subinterfaces

DanG100:


dataplane/dplanerc/interface.go
```diff
@@ -531,18 +539,34 @@ func (ni *Reconciler) reconcile(ctx context.Context, config *oc.Interface) {
 
 	intf := ocInterface{name: config.GetName(), subintf: 0}
 	state := ni.getOrCreateInterface(config.GetName())
-	data, ok := ni.ocInterfaceData[intf]
+	data := ni.ocInterfaceData[intf]
```
guoshiuan, line 554: I think we still need to check whether data is nil before passing it to reconcileEthernet?

guoshiuan (approved):

dataplane/dplanerc/interface.go
```diff
@@ -582,118 +615,188 @@ func (ni *Reconciler) reconcile(ctx context.Context, config *oc.Interface) {
...
+	for idx, subintf := range config.Subinterface {
+		intfRef := ocInterface{
+			name:    config.GetName(),
+			subintf: idx,
+		}
+		data, ok := ni.ocInterfaceData[intfRef]
+		// We have a new interface.
+		if !ok {
+			switch {
+			case config.GetAggregation().GetLagType() != oc.IfAggregate_AggregationType_UNSET:
```
robshakir, line 630: I think `/interfaces/interface/config/type` tells you this more directly, no? If the type is set to `IETFInterfaces_InterfaceType_ieee8023adLag` then this means that this is a LAG.

dataplane/dplanerc/interface.go
```diff
@@ -582,118 +615,188 @@ func (ni *Reconciler) reconcile(ctx context.Context, config *oc.Interface) {
...
+				log.Infof("creating new lag interface: %v", intfRefToDevName(intfRef))
+				if err := ni.createLAG(ctx, intfRef, config.GetAggregation().GetLagType()); err != nil {
+					log.Warningf("failed to create lag: %v", err)
 				}
+			case idx != 0 && subintf.Vlan != nil:
```
robshakir, line 635: There's nothing in OC that tells you that you can't use subinterface index 0 for a VLAN encapsulation (strictly). We tend not to though (by convention) since there are some platforms that have the idea of an interface w/o a subinterface.

dataplane/dplanerc/interface.go
```diff
@@ -582,118 +615,188 @@ func (ni *Reconciler) reconcile(ctx context.Context, config *oc.Interface) {
...
-	}
 
-	// Get all config IPs and their corresponding state IPs (if they exist).
-	for _, addr := range config.GetOrCreateSubinterface(0).GetOrCreateIpv4().Address {
-		pair := &prefixPair{
-			cfgIP: addr.Ip,
-			cfgPL: addr.PrefixLength,
+		// Get all state IPs and their corresponding config IPs (if they exist).
+		var interfacePairs []*prefixPair
+		for _, addr := range state.GetOrCreateSubinterface(idx).GetOrCreateIpv4().Address {
```
robshakir, line 738: Should this just be `Get` vs. `GetOrCreate` to avoid mutating the state data?

dataplane/dplanerc/interface.go
DanG100: is the correct model then that /interfaces/interface == physical port and /interfaces/interface/subinterface == L3 interface

DanG100: or is there a type value i should be using for this. right now i only care about vlan tagged packets getting assigned to different network interface

DanG100: added todo for this

dataplane/dplanerc/interface.go
DanG100: this should be fixed once i migrate this to the new reconciler api

PR #566, opened 2025-04-28
Update gribigo and ygot

DanG100:


robshakir (approved):
Looks like the diff on the protos is caused by a relatively significant difference in protoc version and a change to protoc-gen-go, but please check before submitting.

PR #578, opened 2025-06-05
Implement supervisor switchover functionality

mtr002:


gnoi/gnoi.go
```diff
@@ -338,6 +445,32 @@ func (s *system) IsActiveSupervisor(ctx context.Context, componentName string) (
 	return roleVal == oc.PlatformTypes_ComponentRedundantRole_PRIMARY, nil
 }
 
+// getSupervisorRole validates and returns the active and standby supervisors
+func (s *system) getSupervisorRole(ctx context.Context) (string, string, error) {
```
lifuzhang1108, line 449: Similarly, name the return variables

gnoi/gnoi_test.go
```diff
@@ -491,3 +497,432 @@ func TestComponentReboot(t *testing.T) {
...
+				// Verify response fields
+				if resp.ControlProcessor == nil {
+					t.Error("Expected ControlProcessor in response")
+				}
+				if resp.Version == "" {
+					t.Error("Expected Version in response")
+				}
+			},
+		},
+		"successful-switchover": {
```
lifuzhang1108, line 598: Consider simulate time progression in case we need to extend reboot duration in the future

mtr002: The best way I can think of doing that is with dependency injection (using Time interface). Do you think I should do that in a separate PR?

gnoi/gnoi.go
```diff
@@ -328,8 +329,114 @@ func (s *system) CancelReboot(ctx context.Context, c *spb.CancelRebootRequest) (
...
+		// Return successful response for no-op case
+		return &spb.SwitchControlProcessorResponse{
+			ControlProcessor: r.GetControlProcessor(),
+			Version:          component.GetSoftwareVersion(),
+			Uptime:           0,
+		}, nil
+	}
+
+	switchoverTime := time.Now().UnixNano()
+	if err := fakedevice.SwitchoverSupervisor(ctx, s.c, targetSupervisor, activeSupervisor, switchoverTime); err != nil {
```
marcushines, line 401: so the problem here is all real devices will actually drop connection at this point - basically all they will be able to return is there is pending request to do this so your log message below is kinda not really gonna be able to be true as you are effecitively saying yes we switched over but  you couldn't have based on reality - this has to return first then somehitng in the background has to do the actual switchover

mtr002: Please check the latest commit

PR #588, opened 2025-07-22
Implement packet link qualification

mtr002:


gnmi/fakedevice/fakedevice.go
```diff
@@ -593,3 +595,416 @@ func generateNewPID(ctx context.Context, c *ygnmi.Client, excludePID uint32) (ui
...
+	return nil
+}
+
+// updatePacketStatistics calculates realistic packet statistics for both generators and reflectors
+func updatePacketStatistics(
+	result *LinkQualificationResult,
+	packetConfig *PacketConfiguration,
+	elapsed time.Duration,
+	cfg *configpb.Config,
+) {
```
robshakir, line 966: (style) Same line.

gnoi/gnoi.go
```diff
@@ -745,24 +752,642 @@ func (s *system) getSupervisorRole(ctx context.Context) (activeSupervisor, stand
 	}
 }
 
+type linkQualification struct {
+	plqpb.UnimplementedLinkQualificationServer
+
+	c      *ygnmi.Client
+	config *configpb.Config
+
+	mu sync.RWMutex
```
robshakir, line 761: It's idiomatic to note above this `mu` what it is protecting -- I assume that this is protecting both the `qualifications` and the `historicalResults` maps?

gnoi/gnoi.go
```diff
@@ -745,24 +752,642 @@ func (s *system) getSupervisorRole(ctx context.Context) (activeSupervisor, stand
 	}
 }
 
+type linkQualification struct {
```
robshakir, line 755: I'd suggest to move all of the link qualification code into a new file (`linkqual.go`) within the same package -- otherwise this `gnoi.go` file is going to get somewhat unmanageable.

robshakir (commented):
Will come back and review the remainder when I get a moment -- submitting comments thus far.

For future PRs changes -- smaller changes (~500 lines) are ideal (see the discussion on "smallcls" internally, I can ping you with the link) since they allow for better review velocity, and more detailed review of the code that is there. This is a pretty large change that has a lot of logic in it, so is quite difficult to review.

gnmi/fakedevice/fakedevice.go
```diff
@@ -593,3 +595,416 @@ func generateNewPID(ctx context.Context, c *ygnmi.Client, excludePID uint32) (ui
...
+	if err := executeSetupPhase(ctx, c, interfaceName, timing.SetupDuration); err != nil {
+		return fmt.Errorf("setup phase failed: %w", err)
+	}
+
+	// RUNNING phase
+	if err := checkContextAndUpdate(plqpb.QualificationState_QUALIFICATION_STATE_RUNNING); err != nil {
+		return err
+	}
+	if err := executeRunningPhase(ctx, result, packetConfig, timing, updateCallback, cfg); err != nil {
+		return fmt.Errorf("running phase failed: %w", err)
```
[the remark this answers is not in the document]
mtr002: The error state is already handled by the caller RunPacketLinkQualification function. When executeQualificationStateMachine returns an error, RunPacketLinkQualification sets the qualification state to QUALIFICATION_STATE_ERROR. So we have error state management at the top level rather than having this logic for each phase.

gnmi/fakedevice/fakedevice.go
```diff
@@ -593,3 +595,416 @@ func generateNewPID(ctx context.Context, c *ygnmi.Client, excludePID uint32) (ui
...
+			interfaceSpecs := cfg.GetInterfaces().GetInterface()
+			if len(interfaceSpecs) == 0 {
+				log.Warning("No interfaces found in configuration")
+				return nil
+			}
+
+			log.Infof("Initializing %d network interfaces from configuration", len(interfaceSpecs))
+
+			for _, intfConfig := range interfaceSpecs {
+				intf := &oc.Interface{
```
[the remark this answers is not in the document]
mtr002: Yes! Looking at the code (dataplane/dplanerc/interface.go), you're right—we should add a check for already existing interfaces.
Done.

marcushines (approved):

PR #698, opened 2026-06-12
Add support for handling hop-by-hop packets in Lucius

tejatamboli02:
This diff ensures that Hop-by-Hop packets are no longer prematurely dropped as malformed by the parser. Instead, it allows them into the pipeline where it installs explicit rules to drop unicast Hop-by-Hop packets while permitting multicast Hop-by-Hop packets to continue.
table.go: A WithPriority(priority uint32) method was added to the FlowEntryBuilder. The dataplane uses flow tables to match packets against rules. Because the new Hop-by-Hop rules need to distinguish between multicast and unicast packets, the rules need explicit priorities to ensure the switch evaluates them in the correct order.
switch.go: In the createInvalidPacketFilter function, it uses the new WithPriority API to install specific Hop-by-Hop filtering rules into the IPv6 invalid ingress table (invalidIngressV6Table). It applies 3 rules Existing invalid prefix rule (Priority 20), Rule 1 - Multicast Hop-by-Hop (Priority 10), Rule 2 - Unicast Hop-by-Hop (Priority 5).

### DIFF
--- a/dataplane/forwarding/fwdconfig/table.go
+++ b/dataplane/forwarding/fwdconfig/table.go
@@ -203,7 +203,8 @@ func (b ActionEntryBuilder) set(ed *fwdpb.EntryDesc) {
 
 // FlowEntryBuilder builds flow table entries.
 type FlowEntryBuilder struct {
-	fields []*PacketFieldMaskedBytesBuilder
+	fields   []*PacketFieldMaskedBytesBuilder
+	priority uint32
 }
 
 // FlowEntry creates a new flow entry builder.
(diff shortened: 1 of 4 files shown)


dataplane/forwarding/protocol/ip/ip6.go
```diff
@@ -185,6 +185,10 @@ func makeIP6(frame *frame.Frame) (header, fwdpb.PacketHeaderId, error) {
 		return nil, fwdpb.PacketHeaderId_PACKET_HEADER_ID_NONE, fmt.Errorf("ip6: makeIP6 failed, err %v", err)
 	}
 
+	// Packets with Hop-by-Hop options (Next Header 0) shouldn't be dropped in the parser
+	// because returning an error here treats them as malformed. This prevents them from
+	// reaching the forwarding pipeline where they might need to be punted to the CPU.
```
sonikajindal: Is this just a comment? No code change?

tejatamboli02: yes. I can remove it

sonikajindal: But I thought this is the main issue that the packet was being dropped here as described by the comment?

tejatamboli02: the parser (ip6.go) does not need any change. It just needed change in dataplane tables (switch.go + table.go) to forward those packets instead of dropping them

sonikajindal (approved):

PR #716, opened 2026-08-24
Enable Lucius libsai.so builds using bazel

royyi8:


sonikajindal:
Thanks for this update of Lucius SAI, I have a few questions:

1. What is the difference between *.grpc.pb.h an .h files? I thought they are kept in separate directories for a reason, first is the proto and second is the c++ client stub, so why are we changing the paths? 
2. In all the _grpc.go and grpc_pb.go in proto/sai, I see that we are adding many new attributes, are they from internal SAI or from the open source SAI?
3. Don't we need a sai.patch now?

sonikajindal (approved):

royyi8:
> Thanks for this update of Lucius SAI, I have a few questions:
> 
> 1. What is the difference between *.grpc.pb.h an .h files? I thought they are kept in separate directories for a reason, first is the proto and second is the c++ client stub, so why are we changing the paths?
> 2. In all the _grpc.go and grpc_pb.go in proto/sai, I see that we are adding many new attributes, are they from internal SAI or from the open source SAI?
> 3. Don't we need a sai.patch now?

1. In this PR, we are generating libsai client files (.cc and .h) and the grpc protobuf files(.grpc.pb.h and .grpc.pb.cc) in a single directory: `dataplane/proto/sai`. Both Lucius libsai.so and the go saiserver code is compiled using the generated files in this directory rather than using the static paths in the repo. We consolidated the generated code into a single directory so that we can declare all generated outputs in a single genrule, which makes it easier to resolve header include paths."

2 & 3. 

This block is from the top level MODULE.bazel:

````
    patch_args = ["-p1"],
    patches = ["//patches:sai.patch"],
    sha256 = "94b7a7dd9dbcc46bf14ba9f12b8597e9e9c2069fcb8e383a61cdf6ca172f3511",
    strip_prefix = "SAI-1.15.0",
    urls = ["https://github.com/opencomputeproject/SAI/archive/refs/tags/v1.15.0.tar.gz"],
````

We use open source SAI 1.15 and apply sai.patch. This is the same approach as we do currently except we compile Lucius libsai.so using bazel rather than using makefile.

PR #727, opened 2026-09-15
Changes for UDP to process header and payload correctly

hardikharti:
UDP was terminal. The parse() function was previously calling 'frame.ReadHeader(frame.Len())', consuming the entire remaining frame into udp.header. This left 0 bytes in frame for downstream payload handlers.


dataplane/forwarding/protocol/udp/udp.go
```diff
@@ -165,14 +167,15 @@ func parse(frame *frame.Frame, desc *protocol.Desc) (protocol.Handler, fwdpb.Pac
 	if frame.Len() < udpBytes {
 		return nil, fwdpb.PacketHeaderId_PACKET_HEADER_ID_NONE, fmt.Errorf("udp: parse failed, frame length %v too small to contain a UDP header", frame.Len())
 	}
-	header, err := frame.ReadHeader(frame.Len())
```
sonikajindal, line 168: Is this why we were consuming the whole packet in header?

hardikharti: Yes thats right

sonikajindal (approved):

